PR #18, opened 2021-07-29
Hannah/logging

### LINKED ISSUE #6: Node: Logging
- [x] #16
- [x] #17

### LINKED ISSUE #16: add logging middleware

### LINKED ISSUE #17: allow for environment-dependent logging strategies

hannahle:


server/src/logger/devLogger.ts
```diff
@@ -0,0 +1,33 @@
+import { format, createLogger, transports } from 'Winston';
```
cklamann: I needed this to be lower-case to get the code to compile

cklamann (changes requested):
Looks good! `Winston` import is the only issue I saw on my side.

server/src/logger/devLogger.ts
hannahle: Done!

cklamann (approved):

PR #19, opened 2021-08-03
Hannah/table controls

### LINKED ISSUE #3: React: Table components
- [x] #8
- [x] header cell component with sort controls
- [x] footer cell component with pagination controls
- [x] global search control (does not necessarily have to be part of table component)
- [x] column-specific "advanced-search" control (does not necessarily have to be part of table component)

hannahle:
closes #3 
- Added table sorting, pagination, general and advanced filters
- Replace logos from Material UI with react-icons


react/src/react-table-config.d.ts
```diff
@@ -0,0 +1,124 @@
+/**
+ * This config file is required for hooks to work in Table.tsx.
```
hannahle, line 2: I wasn't able to use some built-in hooks in react-table and followed the steps here: https://github.com/tannerlinsley/react-table/blob/master/TYPESCRIPT.md to fix it

react/src/components/Input.tsx
```diff
@@ -2,8 +2,9 @@ import React from 'react';
 import styled from 'styled-components';
 
 interface InputProps {
-    value: number;
+    value: any;
```
cklamann: I think this can be number or string. We might just default to string and let handler functions take care of conversions if necessary.

react/src/components/Table/ColumnFilter.tsx
```diff
@@ -0,0 +1,28 @@
...
+    setFilter: (columnId: string, filterValue: any) => void;
+    columnId: string;
+}
+
+const SearchInput = styled(Input)`
+    min-height: 30px;
+`;
+
+export const ColumnFilter: React.FC<ColumnFilterProps> = ({ setFilter, columnId }) => {
+    const [value, setValue] = useState('');
```
cklamann: Can you make this component controlled so that it takes `state.filters` as a prop rather than storing the filter value in local state?

react/src/components/Table/GlobalFilters.tsx
```diff
@@ -0,0 +1,33 @@
...
+    filter: any;
+    setFilter: (filterValue: any) => void;
+}
+
+const SearchInput = styled(Input)`
+    margin: ${props => props.theme.space[3]};
+    height: 100%;
+`;
+
+export const GlobalFilter: React.FC<GlobalFilterProps> = ({ filter, setFilter }) => {
```
cklamann, line 16: Can this also be controlled, with local state removed?

cklamann (changes requested):
Looks really good! I made a couple of comments about the inputs -- I'm not sure if it's straightforward to remove local state, but could you please take a look and see if it's possible?

cklamann (approved):

PR #33, opened 2021-08-06
React: Add column grouping feature 

### LINKED ISSUE #28: React: create grouped table columns
The details of the actual groupings are TBD, but for the time being we can group our current fields in the following manner, with the expectation that we'll need to change the configuration down the line.

|field|group|
|---|---|
|af | variation_details|
|alt | core|
|chromosome | core|
|datasetId | case_details|
|dp | case_details|
|end | core|
|ethnicity | case_details|
|phenotypes | case_details|
|ref | core|
|rsId | variation_detail|
|sex | case_details|
|someFakeScore | variation_details|
|start | core|
|zygosity | case_detail|

hannahle:
- Added more columns to table
- Added generic Modal component
- Added column group hiding/unhiding functionality


cklamann (commented):
Looks good! Would it be possible to include the column group names in the header as a second tier? This might require the padding on the table header to be reduced. Something like this:

|       group1         |         group2      | 
|---|---|
| col1 - col2 - col3 | col4 - col5 - col6|

cklamann (approved):

PR #36, opened 2021-08-06
Added chips below search bar

### LINKED ISSUE #35: React: Add chips for multiple sources
- Instead of having a multiselect component, we'd have a list of "chips" that represent the selections.

hannahle:
- Added chip components, which can be useful when we want to have a list of values after user selects multiple options from the combobox.
- Prototype of what it looks like is included in the `VariantQueryPage`. If it looks good, I'll remove it from the file so we can merge to develop.


cklamann (commented):
I think that in the context of the sources this might be something we'd want to implement once we have a few more sources to choose from. For a multiselect, would the chips populate the input box (similar to what MUI does with multiselect/autocomplete)? I'm a bit partial to the checkbox-in-dropdown version to save real-estate if possible, since our input boxes are kind of narrow. If you think it will be better, though, it would be good to see the dropdown implementation

hannahle:
> I think that in the context of the sources this might be something we'd want to implement once we have a few more sources to choose from. For a multiselect, would the chips populate the input box (similar to what MUI does with multiselect/autocomplete)? I'm a bit partial to the checkbox-in-dropdown version to save real-estate if possible, since our input boxes are kind of narrow. If you think it will be better, though, it would be good to see the dropdown implementation

Hmm yes, now that you mention it, a dropdown with checkboxes might be better. I'll try both locally and see which one looks better!

hannahle:
I'm closing this PR because I've merged the branch with #89 .

PR #45, opened 2021-08-12
Hannah/expandable column group

hannahle:
- Make column group headers expandable
- Added individual column filtering to modal


cklamann (approved):
Looks great!

PR #47, opened 2021-08-13
Added CSV export

### LINKED ISSUE #43: React: Download all table data
There should be a button alongside other table controls that gives users the option of downloading table data as a csv into the browser.

hannahle:
- Added an export button that downloads data from all currently visible columns


cklamann (changes requested):
There are a few wrinkles here that we'll have to consider, the biggest being the accessor functions. For instance, the display value for the `phenotypes` column is a concatenation of phenotype ids pulled from the `phenotypicFeatures` field and not the raw array of objects or any field called `phenotypes`. The download should have the same format, which means the accessor function stipulating the format of `phenotypes` should be available to the download code, or else it should be called on the data before being passed to the download code. A second thing we may need are additional "download accessor functions" for columns that need to be formatted differently in csv form. For instance, the `contact` field will render a button in the table but should print an email address in the csv. We'll need to do some research to see if/how this fits in with the react-table API and whether we'll need some additional config for specifying csv formatting for certain fields beyond what we can do with the react-table column fields.

cklamann (approved):

PR #54, opened 2021-08-24
React: Add error handlers and display components

### LINKED ISSUE #49: React: Error Handling
Errors can arise at various points in the query resolution process. The ssmp-server *should* (though may not currently) catch and forward client-node HTTP errors to the front end (in the `errors` property of the return schema). The ssmp-server might also generate HTTP errors of its own (i.e, server-side exceptions, including validation exceptions). Finally, there might be graphql errors returned in the `errors` property of the top-level query payload. The front-end should have separate alerts and recovery logic for each of these errors, since they indicate to failures at different points of the application with different implications for the user. This issue can be closed once we:

- [ ] Create error handlers and display components for SSMP-generated HTTP errors.
- [ ] Create error handlers and display components for client-node-generated HTTP errors (including timeouts).
- [ ] Create error handlers and display components for graphql errors.

hannahle:


cklamann (approved):
Looks great!

PR #56, opened 2021-08-31
Node: Add basic request logging

### LINKED ISSUE #55: Node: Log all requests
Each time a request is made, the server should log the query and the identity of the requester.  This will probably require the backend to call the `userinfo` keycloak endpoint on each request, ideally caching the result with the token as the key.

cklamann:
Closes #55


hannahle (approved):
LGTM!

PR #61, opened 2021-09-15
Customize keycloak login

### LINKED ISSUE #57: Keycloak: Customize login  page
We should customize the keycloak login page to use as much of our theme as possible.

hannahle:
- /opt/jboss/keycloak has a default theme folder that contains the base keycloak theme. I created a local theme folder that contains static css files and attached that to the keycloak's volume to provide keycloak with an extra theme. 
- This theme is then loaded automatically to our server in the bootstrap-keycloak.sh file (See added line).
- For development purposes only, the `disabled-theme-cache.cli` disables theme caching and allows for theme code update to be reflected when the page is refreshed, as opposed to having to restart the Docker container. This should not be added to the production build for keycloak.


cklamann (approved):
LGTM!

PR #62, opened 2021-09-17
Docker: update env vars

cklamann:


hannahle (approved):
Looks good to me.

It didn't work before mainly because I was working out of UofT today and the network was extremely slow... But otherwise works fine!

PR #67, opened 2021-09-21
React: enable gene search autocomplete

### LINKED ISSUE #13: React: Enable Gene Search

- [x] #15
- [x] #14
    - autocomplete from gene name: https://www.ebi.ac.uk/ebisearch/ws/rest/ensembl/autocomplete?term=apo 
    - get ens id from name:https://www.ebi.ac.uk/ebisearch/ws/rest/ensembl_gene?query=apoe&size=1000
    - get region from ens id: http://rest.ensembl.org/lookup/id/ENSG00000130203
- [x] add gene search to graphql schema and integrate with test data
- [x] ~~investigate/address circumstances when autocomplete api returns no results for valid query (easiest fix is just to allow user to enter arbitrary gene identifier and submit)~~ UPDATE: we changed endpoints and this issue should be resolved

cklamann:
closes #13


react/src/pages/VariantQueryPage.tsx
```diff
@@ -145,10 +146,12 @@ const VariantQueryPage: React.FC<{}> = () => {
...
                         </Typography>
-                        <Input
-                            disabled={!!queryOptionsForm.ensemblId.value}
-                            onChange={e => updateQueryOptionsForm('gene')(e.currentTarget.value)}
-                            value={queryOptionsForm.gene.value}
+                        <GeneSearch
+                            selectedGene={queryOptionsForm.gene.value}
+                            onSelect={geneOption => {
+                                updateQueryOptionsForm('gene')(geneOption.value.name);
+                                updateQueryOptionsForm('ensemblId')(geneOption.value.ensemblId);
```
hannahle, line 154: Say user has selected a gene (like APOE), the ensmbl id field will get autopopulated. When user edits the ensmbl id, I think we should clear out the gene name automatically as well because now the ensembl id and the gene name don't represent the same gene anymore.

react/src/components/GeneSearch.tsx
```diff
@@ -1,79 +1,66 @@
...
-                    (s: { suggestion: string }, i: number) => ({
-                        value: s.suggestion.toUpperCase(),
-                        id: i,
-                        label: s.suggestion.toUpperCase(),
-                    })
-                )
+                (autocompleteResults.autocompleteResults.hits || []).map((hit, i) => ({
+                    value: {
+                        name: hit.symbol.toUpperCase(),
+                        ensemblId: hit.ensembl?.gene,
```
hannahle: Because the ensemblId can be `undefined` here, the program throws an error at this line [here](https://github.com/ccmbioinfo/ssmp/blob/4b92a0f8961a58ec7240cd4241dcd508f53bf299/react/src/pages/VariantQueryPage.tsx#L33). An example is when the gene APOC4-APOC2 is selected, its ensemblId is undefined. We probably would want to filter out undefined results.

hannahle: I think we should also improve the way we're validating ensembl id as well. Right now, the logic is to check if the ensembl id starts with `ENSG00`, but a more robust mechanism might be to check the ensembl id against an endpoint (Maybe this one here: https://mygene.info/v3/api#/annotation/get_gene__geneid_).

hannahle (approved):
LGTM!

PR #74, opened 2021-09-23
React: Fixed table header

### LINKED ISSUE #71: React: Sticky table headers
We'll probably want the table headers to stay in place as the user scrolls down, especially if hundreds of results are visible at once.

hannahle:
- Added sticky table header 
- Changed size of button so it fits to the column


cklamann (approved):
LGTM!

PR #75, opened 2021-09-27
React: Resolve fixed table header causing collapsed cells error

### LINKED ISSUE #71: React: Sticky table headers

hannahle:


react/src/components/Table/Table.styles.tsx
```diff
@@ -42,6 +42,11 @@ export const Styles = styled.div`
                 border-right: 0;
             }
         }
+
+        tbody {
+            max-height: '60vh';
```
cklamann: I don't think the style values need to be quoted. It also seems that the "sticky header" doesn't work when the height is applied directly to `tbody`. I'm not sure, but I think this has to do with the default `display` value of `tbody`, which is `table-row-group`. By changing it to `block` in the class, I was able see the desired behavior. There is also the minor issue of the vertical scrollbar pushing cells to the left, which, especially at the right side of the table, skews column alignment. I was able to fix it by adding `margin-right: -15px` to this class, but there might be a better solution.

cklamann (changes requested):

cklamann (approved):

PR #78, opened 2021-09-29
React: create select component and add to query tools

### LINKED ISSUE #77: React: Create select element and add to form widgets
We want to replace the native `select` and `option` elements with something that gives us more control over styling and use it to display certain queryable fields (genome assembly, source). A good amount of this can probably be accomplished by breaking up the `combobox` component.

cklamann:
closes #77
- Update combobox to be either static (like a `select` element) or dynamic ("autocomplete")
- update styles


hannahle (approved):
I added one line of code to remove the front padding in the Gene Search box; otherwise looks good to me!

PR #89, opened 2021-10-19
React: Added sign comparison for number fields in Advanced Filters and created dynamic dropdown

### LINKED ISSUE #87: React: Improve filter options for Advanced Filters section
The advanced filters currently only have broad text search without strict validation of what user enters. Depending on the data type, we should add more useful filter expressions (such as less than / greater than). The filter for Contact should also be removed. 

Add debounce to filters

### LINKED ISSUE #94: React: create (potentially) dynamic dropdown filters
For certain fields, there should be a dropdown whose options are static or supplied dynamically from the results. These should be fields where the number of possible results is limited or known ahead of time or a small number of unique values is expected, allowing users to both filter results from the options and get a quick look at the range of unique values for that variable. The fields `sex`, `zygosity`, and `consequence` seem like good starting places.

hannahle:
- Added sign dropdown for the `start` and `end` field. 
- Added Chromosome box dropdown
- Replaced `styled-components` with `styled-components/macro` for easier debugging (`styled-components` usually generates obscured names in the HTML tree when inspecting the browser. `macro` would allow it to generate more human-readable names, like `StyledInput` to help with debugging).
- Added InputStartAdornment to Input box
- Added dynamic options for dropdown
- Added searchability for dropdown with multi-select
- Enabled multi-select for dropdown with checkboxes


cklamann (commented):
I'm seeing some odd behavior with the `start` and `end` filters, specifically with values reappearing in the boxes after deletion. Can you try entering and deleting some dummy values and see if you observe the same thing?

react/src/components/Table/TableFilter/NumberRangeFilter.tsx
```diff
@@ -0,0 +1,141 @@
...
+    };
+
+    const filterValue = filters.find(f => f.id === columnId)?.value as Array<number>;
+    console.log('filter value', filterValue);
+    const comparison = filterComparison[columnId];
+
+    const resolveComparisonComponent = () => {
+        if (comparison.less) {
+            return (
+                <Input
```
cklamann: There's a good amount of shared props between the Input components in this module. Do you think it's possible to consolidate them?

hannahle: Yeah, definitely. We can pass the common variables to a function. Let me try it out!

hannahle:
Converting this back to draft to also add dynamic filter option for SelectionFilter.

cklamann (approved):
LGTM!

PR #97, opened 2021-10-29
React: Added ellipsis text overflow with popover for Alt, Ref, and Contact

### LINKED ISSUE #95: React: handle column overflow
It's possible that we will get very large strings in some fields (esp alt/ref for large indels). We should accommodate these by styling the column fields with `overflow: ellipsis` and providing a popover for users to view the full result.

hannahle:
Closes #95


.gitignore
```diff
@@ -8,7 +8,11 @@ yarn-debug.log*
 yarn-error.log*
 
 .devcontainer
+<<<<<<< HEAD
```
cklamann: Incomplete merge

react/src/components/Table/CellPopover.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { FlattenedQueryResponse } from './Table';
+
+interface CellPopoverProps<T> {
+    state: T;
+    id: keyof T;
+}
+
+const Container = styled(props => <Background {...props} />)`
+    padding: ${props => props.theme.space[1]} ${props => props.theme.space[5]};
+    width: 300px;
```
cklamann, line 16: I'm still seeing container overflow for very long strings (like some in PRSS3) -- removing this width setting seems to solve it but I'm not sure if there are other implications.

cklamann (changes requested):

cklamann (approved):
LGTM!

PR #99, opened 2021-11-02
Devops/Misc: update dev compose, add sources as env var, clean up annotation fetch a

cklamann:
- remove pubsub
- clean up dev compose and some stray env vars
- increase timeout and max buffer size for annotation fetch


hannahle (approved):
LGTM!

PR #115, opened 2021-11-09
React: Debounce filter input to limit unnecessary table rerendering 

### LINKED ISSUE #113: React: Add debounce to advanced filters
Not having debounce can potentially lead to performance issue and misleading error messages when user has not finished typing. Adding a debounce functionality can solve the mentioned issues.

hannahle:
- Closes #113
- Added `useDebounce` hook for debouncing the filter function by react-table. 
- The main idea is we have to store a local state `input` and the debounced state `debouncedInput` would be set after `timeout=500`. `useEffect` listens to changes in `debouncedInput` and calls the `setFilter` function on the Table component.


react/src/components/Table/TableFilter/ColumnFilter.tsx
```diff
@@ -27,6 +28,17 @@ export const ColumnFilter: React.FC<ColumnFilterProps> = ({
     setFilter,
     type,
 }) => {
+    const [debouncedInput, input, setInput] = useDebounce<string | undefined>('', 500);
```
cklamann, line 31: This seems to be working as expected on the regular text inputs. However, the number range filters seem to be functioning as before -- input is passed straight to the filters w/o debounce.

hannahle: That's right. It's because I didn't pass in the debounce input to the number range component. I can add that.

hannahle:
Closing this PR because #134 is more up-to-date and re-uses `useAsyncDebounce` by react-table.

PR #118, opened 2021-11-09
Mongo: bring back mongo as gnomad annotation storage

cklamann:


README.md
```diff
@@ -60,3 +60,23 @@ docker exec -i <keycloak-container-name> bash /usr/scripts/bootstrap-keycloak.sh
...
+   --file=<filename>.csv \
+   -u <username> --password=<pass> --authenticationDatabase=admin
+```
+
+Then make sure to create the following indexes:
+
+```
+db.annotations.createIndex({"pos": 1})
+db.annotations.createIndex({"assembly":1})
+db.annotations.createIndex({"alt": 1, "chrom": 1, "pos": 1, "ref": 1 })
```
hannahle: This can also be run with a single command as `db.annotations.createIndexes([
{"pos": 1}, {"assembly":1}, {"alt": 1, "chrom": 1, "pos": 1, "ref": 1 }
])`

react/src/types.ts
```diff
@@ -2,14 +2,15 @@ import { Maybe } from 'graphql/jsutils/Maybe';
 
 /* typescript types that map to graphql types, should be updated whenever schema is updated -- note that these are coming from our own server now */
 export interface VariantResponseInfoFields {
+    af?: Maybe<number>;
     aaAlt?: Maybe<string>;
     aaPos?: Maybe<string>;
     aaRef?: Maybe<string>;
-    cdna?: Maybe<string>;
+    cdna?: Maybe<number>;
```
hannahle: I'm not sure if `cdna` is always of type `number`. At least for gnomad, if you use panda to get back unique values for cdna, I recall sometimes seeing 21312-12312 which is a string.

hannahle: The rest of the annotation logic looks good to me!

hannahle (approved):
I think it's ready!

PR #130, opened 2021-11-15
React: make multiselect a custom filter type

### LINKED ISSUE #129: Bug: Not all results are shown when multiple options are selected
When we select >1 option for the combobox in Advanced Filters, no results are returned, although the data does exist and would filter if only one option is selected.

cklamann:
- Closes #129


hannahle (approved):
LGTM!

Before, we didn't need `filterTypes` because we were using the original React Table's `includesMany` where the `filterValue` is a list of string versus a string, and the logic of what to do with that list of string is in `ColumnFilter`. Now that we move the logic to the column config, we no longer have the built-in filtering for multiple values, hence the need for `filterType`.

PR #131, opened 2021-11-15
React: move computed values out of accessors and update types

### LINKED ISSUE #125: React: csv export should not contain html or javascript objects
Currently `Cell` and other display components are included in the csv export. We should export only the data from the results.

cklamann:
- Closes #125 
- Separate flattened query response type from column value type
- consolidate formatting code
- update csv download logic


react/src/components/Table/Table.tsx
```diff
@@ -520,8 +514,8 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         variant="primary"
                         onClick={() => {
                             downloadCsv(
-                                formatDataForCsv(rows),
-                                visibleColumns.map(c => c.id)
+                                rows.map(r => r.values as ResultTableColumns),
```
hannahle, line 517: I think we'd need to filter out the dummy columns (`emtpyVariationDetails`, `emptyCaseDetails`) as well so user doesn't have those columns in the CSV when they are expanded.
Screen Shot 2021-11-18 at 12 43 01 PM

hannahle (approved):
LGTM!

PR #135, opened 2021-11-24
React: add metadata on headers' hover

### LINKED ISSUE #126: React: show column info on hover
Each table column header should display metadata about the column's contents on hover. The metadata may change depending on the results, so the component should use a accessor function that takes the full results set as an argument (for now)

hannahle:
- Created a tooltip component that shows metadata on hover


cklamann (approved):
:rocket:

PR #136, opened 2021-11-25
React: Added PhenotypeViewer component

### LINKED ISSUE #128: React: create phenotype viewer component
API payloads may contain many phenotype entries. We should have a component in the table that allows users a quick view of this data, since it will probably be too large to view in a single table cell. One idea is to display only the count of reported phenotypes in the cell and then show the details in a popover. Ease of use could also be improved by adding a multiselect filter to the phenotypes field so that users can get a quick overview of the set of phenotypes returned.

hannahle:


cklamann (approved):
LGTM!

PR #137, opened 2021-11-26
Node: handle 37/38 region in autocomplete, abort cadd on sizes > 600kbp

cklamann:


react/src/apollo/hooks/useFetchAutocompleteQuery.ts
```diff
@@ -23,6 +22,12 @@ const autocompleteQuery = gql`
                     start
                     strand
```
hannahle: Can be removed.

hannahle (commented):
The rest looks good to me!

hannahle (approved):

PR #143, opened 2021-12-01
React: Cleaned up react-table

### LINKED ISSUE #132: React: Refactor Table.tsx
Table.tsx is getting harder to navigate due to the larger codebase. 

There are several things we can do to improve readability and usability going forward:
- [x] Refactor TableRows, TableCells, Header, Footer into its individual functional component.
- [x] Separate utility functions into a separate file. 
- [x] Separate logic of checking whether a column is a header/ expanded, etc. to a separate file.

### LINKED ISSUE #144: React: move non-hooks out of hooks directory
resolveAssembly, downloadcsv should go in utils (making issue so i don't forget :slightly_smiling_face:)

hannahle:
- Moved utility functions to a separate file
- Cleaned up some small Typescript for column ids and setFilter function (specifically re-using types of react-table versus redefining them in props ourselves)
- Moved large components, such as Footer or Advanced Filters into a separate file to avoid cluttering react-table.
- Before, there were some list of column ids declared, such as `dummyColumns`, etc. These are now moved inside the column config to avoid using `useMemo` to manually declare these columns altogether.


cklamann (approved):
Looks great!

PR #152, opened 2021-12-09
Scripts: add demo scripts and Dockerfile for dynamic liftover

cklamann:


scripts/liftover/liftover.js
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env node
```
hannahle, line 1: I remember you mentioning at some point that not all regions can be lifted over (?). I assume that would be a separate check outside of this liftover script right?

cklamann: The liftOver command takes paths to four files as arguments: the input file (BED format), a chain file (the mappings), a file where the lifted coordinates will be written, and a file where the unliftable coordinates will be written. So this script will actually account coordinates that cannot be lifted over, though they are currently not being tagged as such (though they should).

PR #158, opened 2021-12-14
Node: added payload from PT /rest/patients/fetch?id=<id> to individual info fields 

### LINKED ISSUE #157: Node: Create adapters and resolvers for requests to the PT patients endpoint
Currently, we're only querying PT's variant matching endpoint, which doesn't contain individual cases' info, such as geographic origin, ethnicity, and so on. We'd need to add the patient endpoint query to get such information.

hannahle:


kevinlul (commented):
Is the schema for the PhenoTips endpoint documented or based on inspection?

hannahle:
> Is the schema for the PhenoTips endpoint documented or based on inspection?

It's based on Alina's response to my email on what the complete response should be. What's unclear is what is required versus not in each interface. Now you brought this up I should probably change the type from required for most fields to _not_ required.

server/src/resolvers/getVariantsResolver/adapters/g4rdAdapter.ts
```diff
@@ -68,7 +71,7 @@ const getG4rdNodeQuery = async ({
   const url = `${process.env.G4RD_URL}/rest/variants/match`;
   const { position, ...gene } = geneInput;
   try {
-    G4RDNodeQueryResponse = await axios.post<G4RDQueryResult>(
+    G4RDVariantQueryResponse = await axios.post<G4RDVariantQueryResult>(
```
kevinlul, line 76: I don't have much too say on this, but maybe keep local variables camelCase to conform to convention and avoid confusion with types, interfaces, and classes.

kevinlul:
👁️

kevinlul:
👀

PR #160, opened 2021-12-16
Node: replace tabix child processes with tabix Javascript wrapper 

### LINKED ISSUE #124: Node: make CADD tabix query in javascript
As the app matures, we may want to think about dropping the `tabix` child process in favor of a solution in JavaScript. There are several libraries that have done this and from which we can draw inspiration, such as [bioldalliance](https://github.com/dasmoth/dalliance/blob/master/js/tabix.js).

hannahle:


server/package.json
```diff
@@ -59,6 +62,7 @@
     "mongodb-memory-server": "^7.4.3",
     "mongoose": "^6.0.8",
     "node-cache": "^5.1.2",
+    "node-fetch": "2",
```
kevinlul: Is this to avoid the ESM-only node-fetch 3 upgrade?

server/package.json
```diff
@@ -70,5 +74,6 @@
     "mongodbMemoryServer": {
       "debug": "1"
     }
-  }
+  },
+  "type": "module"
```
kevinlul: This changes this project from CJS to ESM. Is that supposed be part of the scope of this change?

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
```diff
@@ -1,86 +1,91 @@
...
-  Alt: 'alt',
-  Consequence: 'consequence',
-  oAA: 'aaRef',
-  nAA: 'aaAlt',
-  FeatureID: 'transcript',
-  cDNApos: 'cdna',
-  protPos: 'aaPos',
-};
+  const tbiIndexed = new TabixIndexedFile({
+    filehandle: new RemoteFile(annotationUrl, { fetch: fetch as Fetcher }),
```
kevinlul, line 30: Are you redownloading this dependency every time Tabix is used?

server/package.json
hannahle: Nope - this was some experimental code that should be removed. I pushed up a new commit that reverted it back to CJS.

server/package.json
hannahle: New commit should revert it back to ESM only!

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
hannahle: The file is >300GB, but the query happens instantly, so no, I don't think the file specified with the `annotationUrl` is being downloaded every single time. I'm not entirely sure how tabix works under the hood, but according to the documentation [here](http://www.htslib.org/doc/tabix.html#:~:text=Last%20but%20not%20the%20least,without%20downloading%20the%20entire%20file.):

> Last but not the least, tabix supports remote data retrieval. One can put the data file and the index at an FTP or HTTP server, and other users or even web services will be able to get a slice without downloading the entire file.

kevinlul: Okay, it's using HTTP byte-range headers to fetch chunks of it then.

kevinlul:
So the project is CJS, but an ESM-only node-fetch was imported?

hannahle:
> So the project is CJS, but an ESM-only node-fetch was imported?

Yes, this was the case. I also realized upon restarting the app that this is not allowed. I removed the node-fetch package altogether and replaced it with cross-fetch, which was already installed.

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
```diff
@@ -1,86 +1,88 @@
...
+      .split('\t')
+      .filter((a, i) => headersIndex.some(headerId => Number(headerId) === i));
+    return Object.fromEntries(
+      headers.map((h, i) => [h, annotationColumns[i]])
+    ) as unknown as CaddAnnotation;
+  });
 
-  return `tabix -h  ${annotationUrl} ${indexPath} ${position} | awk 'NR!=1{print $1,$2,$3,$4,$8,$17,$18,$20,$25,$29}'`;
+  console.log(result);
+  return result;
```
kevinlul: Can you explain what this function is supposed to do?

server/src/resolvers/getVariantsResolver/utils/resolveChromosome.ts
```diff
@@ -0,0 +1,14 @@
+const resolveChromosome = (position: string) => {
+  const chromosome = position.match(/^[^:]*[^ :]/gm)?.join('');
+  const coordinates = position.match(/\w[^:]*$/gm)?.join('');
+  const end = coordinates?.match(/\w[^-]*$/gm);
+  const start = coordinates?.match(/^[^-]*[^ -]/gm);
```
kevinlul: Were these expressions retrieved from elsewhere or did you write them yourself?

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
hannahle: The tabix JS wrapper returns a list of strings, each string represents a line in the annotation tsv from `ANNOTATION_URL`, i.e. a variant annotation. It looks something like this: 

`19\t12312321\t23432\t....`

We know what headers correspond to what values based on CADD's documentation [here](https://cadd.gs.washington.edu/static/ReleaseNotes_CADD_v1.6.pdf).

The goal of the function is to take in a list of strings and return a list of objects of type `CaddAnnotation`.

server/src/resolvers/getVariantsResolver/utils/resolveChromosome.ts
hannahle: The last two were written by me once I understand what the individual components like /, ^, [], * do.

The first two I was able to get by consulting this Stackoverflow here: https://stackoverflow.com/questions/3858519/regular-expression-to-match-everything-that-appears-before-a-colon

(Hopefully one day I will be able to write so much regex that I can pull them out of my brain without looking at anything...)

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
hannahle: I will add more detailed comments to the code.

server/src/resolvers/getVariantsResolver/utils/resolveChromosome.ts
kevinlul: Yeah, this is my main concern with regex, which is that it becomes incomprehensible even to the original writer some time after it was written. Often, they aren't even the right tool for the job because of their complexity and runtime (and risks of catastrophic backtracking). Thus I would recommend against them if the objective is achievable with more understandable code; for example, everything before a colon can also be obtained from `.split(":")`.

hannahle: That's a good point. I'll modify the code with split.

server/src/utils/typedObject.ts
```diff
@@ -0,0 +1,3 @@
+export function fromEntries<T>(entries: [keyof T, T[keyof T]][]): T {
+  return entries.reduce((acc, [key, value]) => ({ ...acc, [key]: value }), <T>{});
+}
```
kevinlul: Did Object.fromEntries<T> not work? I don't believe you need to write this.

server/src/resolvers/getVariantsResolver/utils/resolveChromosome.ts
```diff
@@ -0,0 +1,14 @@
+const resolveChromosome = (position: string) => {
+  const chromosome = position.split(':')[0];
+  const coordinates = position.split(':')[1];
+  const end = coordinates.split('-')[0];
+  const start = coordinates.split('-')[1];
```
kevinlul: If it is actually this simple (that you can just split on these tokens assuming exactly one of them is in `position`), then you can also use a destructuring assignment to not split the same string twice:
```suggestion
  const [chromosome, coordinates] = position.split(':');
  const [end, start] = coordinates.split('-');
```

server/src/utils/typedObject.ts
hannahle: No it didn't really work. Object.fromEntries < T > casts the value to whatever key to T.

kevinlul: I see, then you do need to cast the return of that native function then, I misread the type signature.

PR #167, opened 2022-01-06
React: add custom filename when exporting CSV data

### LINKED ISSUE #164: React: Customize names of CSV exports
Currently, all CSV exports have `ssmp.csv` as the default name, which is uninformative. We should have a textbox where user can answer the name of the file they're downloading.

hannahle:


react/src/utils/downloadCsv.tsx
```diff
@@ -16,13 +16,16 @@ const objectToCsv = <T extends object>(data: T[]) => {
     return csvRows.join('\n');
 };
 
-const download = (data: BlobPart) => {
+const formatFileName = (filename: string) =>
+    filename.includes('.csv') ? filename : `${filename}.csv`;
```
kevinlul: Should use `endsWith`

react/src/components/Table/DownloadModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+                    setFilename('');
+                }}
+                title="Save Variants"
+                footer="Download CSV"
+                onClick={() =>
+                    downloadCsv(
+                        rows.map(r => r.values as T),
+                        visibleColumns
+                            .filter(c => c.id && !c.id.match(/^empty/i))
+                            .map(c => c.id as keyof T),
```
kevinlul: Do you still need the typecasts when you've typed the `rows` argument as `Row<T>[]`?

hannahle: I changed the types to IdType (which is what react-table uses for its keyof T | string) type. We'd still need to cast it as keyof T in `downloadCsv.ts` but I think it's cleaner.

react/src/utils/downloadCsv.tsx
```diff
@@ -16,13 +17,16 @@ const objectToCsv = <T extends object>(data: T[]) => {
...
     const blob = new Blob([data], { type: 'text/csv' });
     const url = window.URL.createObjectURL(blob);
     const a = document.createElement('a');
     a.setAttribute('hidden', '');
     a.setAttribute('href', url);
-    a.setAttribute('download', 'ssmp.csv');
+    a.setAttribute('download', formatFileName(filename));
     document.body.appendChild(a);
     a.click();
     document.body.removeChild(a);
```
kevinlul: I know this code was already here rather than modified by the PR, but in Stager we don't even have to append the element to the body or set the a as hidden. The blob URL also isn't freed after the click.

PR #169, opened 2022-01-07
React: fix assembly labels and gene search 

### LINKED ISSUE #165: React: fix assembly ID display label

hannahle:
- Assembly ID selection now reflects the full form of the reference genome (GChr38/37 instead of 38, 37).
- Before, if the user types in the same gene name as before, no autocomplete results would be shown in the dropdown. Adding `geneName` as a dependency to the `useEffect` hook responsible for setting options in the dropdown fixes this issue.

### DIFF
--- a/react/src/components/GeneSearch.tsx
+++ b/react/src/components/GeneSearch.tsx
@@ -1,10 +1,30 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useAsyncDebounce } from 'react-table';
 import { useFetchAutocompleteQuery } from '../apollo/hooks';
 import { AssemblyId } from '../types';
 import ComboBox from './ComboBox';
 import { SelectableListItem } from './SelectableList';
 
+interface AutocompleteResults {
+    autocompleteResults: {
+        hits: {
+            symbol: string;
+            ensembl: {
+                gene: string;
+            };
+            genomic_pos: {
+                chr: string;
+                end: number;
+                start: number;
+            };
+            genomic_pos_hg19: {
+                chr: string;
+                end: number;
+                start: number;
+            };
+        }[];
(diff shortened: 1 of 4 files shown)


react/src/components/GeneSearch.tsx
```diff
@@ -43,6 +63,22 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
 
     const debouncedAutocompleteFetch = useAsyncDebounce(fetchAutocompleteResults, 500);
 
+    const formatAutocompleteOptions = useCallback(
+        (autocompleteResults: AutocompleteResults) =>
+            (autocompleteResults.autocompleteResults.hits || [])
+                .filter(hit => !!hit.ensembl?.gene && !!hit.genomic_pos && !!hit.genomic_pos_hg19)
```
kevinlul, line 60: Guessing by this that most fields are actually optional for `AutocompleteResults`?

server/src/resolvers/getVariantsResolver/utils/resolveAssembly.ts
```diff
@@ -1,11 +1,6 @@
...
-  } else if (assembly.includes('38')) {
-    return '38';
-  } else {
-    return '';
-  }
-};
+import { AssemblyId } from '../../../types';
+
+const resolveAssembly: (assembly: string) => AssemblyId = assembly =>
+  assembly.includes('38') ? '38' : '37';
```
kevinlul: Do you have enforce module boundary typings on for this codebase? If not you might be able to make this less confusing by just typing the argument and adding `as const` after the return.

react/src/components/GeneSearch.tsx
hannahle: ah yes good catch.

PR #177, opened 2022-01-19
React: modified gene search to display multiple ensembl IDs per gene name

### LINKED ISSUE #175: React: modify dropdown for gene name with multiple ensembl IDs
Currently, our dropdown only supports 1-to-1 relationship for 1 gene name that corresponds to 1 ensembl ID. There genes, such as ABHD16A, that has multiple ensembl IDs. We should give users warning about genes with multiple IDs and prompts them to choose the appropriate ID.

hannahle:


react/src/components/GeneSearch.tsx
```diff
@@ -40,6 +42,9 @@ interface GeneSearchProps {
     onChange: (geneName: string) => void;
 }
 
+/* Typeguard */
+const isMultipleGenes = (arg: Gene | Gene[]): arg is Gene[] => (arg as Gene[]).length >= 0;
```
kevinlul: `Array.isArray` is probably a better check

react/src/components/GeneSearch.tsx
```diff
@@ -57,16 +62,35 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
...
+                            name: hit.symbol.toUpperCase(),
+                            ensemblId: e.gene,
+                            position: getPosition(assembly, {
+                                genomic_pos: hit.genomic_pos[eid],
+                                genomic_pos_hg19: hit.genomic_pos_hg19[eid],
+                            }),
+                        },
+                        id: i + eid,
+                        label: isMultipleGenes(hit.ensembl) ? e.gene : hit.symbol.toUpperCase(),
+                    }))
```
kevinlul: You could do this in the prior map call to only loop through the hits once, unless you plan to reuse the separate functions.

react/src/components/GeneSearch.tsx
hannahle: Good point!

react/src/components/GeneSearch.tsx
hannahle: Sounds good!

PR #189, opened 2022-02-08
React: Standardize the display for Zygosity column

### LINKED ISSUE #179: React: display one row per variant and expand when clicking on Case Details
- [x] Standardize the display for **Zygosity** column. Currently, we have mixed texts (such as "het", "Het", "Heterozygous", "heterozygous", etc.) coming from different sources. We should have a check that if the text contains "het" or "hom", display it as "Heterozygous" and "Homozygous" on the table, respectively. For other data that doesn't contain "het" or "hom", leave it as is. This can be accomplished using the `accessor` property in the column configuration of `react-table`. `Accessor` basically allows us to control what texts are being displayed, without changing the value on the backend directly (see more detailed documentation [here](https://react-table.tanstack.com/docs/api/useTable)). 

- [x] Add two columns to **Variant Details** group - **Heterozygous Count** and **Homozygous Count**. Each column represents how many heterozygous traits there are for a given variant. 

- [x] Currently, there are multiple patient cases for a single variant. When only **Variant Details** and **Variant** column groups are expanded, the table contains a list of repetitive rows of the same information. This makes the table look bulky. We only want to display one row for variant when **Case Details** are collapsed
 
- [x] Add alternating colors for the rows.

jennyziyi-xu:
If the `zygosity` field contains "het" or "hom" in lower case or upper case, display it as "Heterozygous" and "Homozygous" on the table, respectively. 
For other data that doesn't contain "het" or "hom", leave it as is.

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -193,7 +193,12 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         width: 70,
                     },
                     {
-                        accessor: 'zygosity',
+                        accessor: state =>
+                            state.zygosity?.toLowerCase().includes('het')
+                                ? 'Heterozygous'
+                                : state.zygosity?.toLowerCase().includes('hom')
+                                ? 'Homozygous'
+                                : state.zygosity,
                         filter: 'multiSelect',
                         id: 'zygosity',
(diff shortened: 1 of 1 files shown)


hannahle (approved):
LGTM!

PR #191, opened 2022-02-09
React: added styled-components/macro to enable more descriptive class names 

### LINKED ISSUE #190: React: Configure styled-components to display more helpful frontend component names

hannahle:
Currently, when debugging `styled-components`, we have randomly generated class names when trying to inspect with the Developer Tools, which can be confusing and unhelpful. `styled-components/macro` plug-in allows for generation of more descriptive class names, such as `StyledButton_welkjdalks` instead of `welkjdalks` like before.

### DIFF
--- a/react/.babelrc
+++ b/react/.babelrc
@@ -1,3 +1,0 @@
-{
-    "plugins": ["babel-plugin-styled-components"]
-}
--- a/react/.babelrc
+++ b/react/.babelrc
@@ -1,0 +1,4 @@
+{
+    "plugins": ["macros"]
+  }
+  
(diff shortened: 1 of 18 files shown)


react/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+    "plugins": ["babel-plugin-styled-components"]
+}
```
kevinlul: I recommend putting this in a "babel" key in package.json so the subproject root doesn't get cluttered with dotfiles. See https://babeljs.io/docs/en/configuration#packagejson

jennyziyi-xu (approved):
LGTM!

PR #196, opened 2022-02-15
React: Added filters popover for each column header

### LINKED ISSUE #183: React: added filter to column headers
Currently, we have the "Advanced Filters" section for filtering certain columns. However, because the number of columns is large, it is difficult for our collaborators to find the right filter that they want. 

A more intuitive user experience would be to have a filter icon next to each Header. Upon clicking on the icon, there would be a popup where user can enter some text or number to filter. The design of the field in each popup can be similar to the existing Advanced Filter.

hannahle:
- Before, we had a section for advanced filters. However, users have expressed that this is hard to use due to the large number of columns available. 
- Now, after user clicks on the filter icon next to each header, an advanced filter box would pop up. 

Other changes include:
- Added a Chip component to keep track of active filters. 
- Modified different filter component codes to keep track of filter values after component has unmounted.

### DIFF
--- a/react/src/components/Chip.tsx
+++ b/react/src/components/Chip.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import { BsX } from 'react-icons/bs';
+import styled from 'styled-components/macro';
+import { Button, Flex, Typography } from './index';
+
+export const ChipContainer = styled(props => <Flex {...props} />)`
+    border-radius: ${props => props.theme.radii.round};
+    background: ${props => props.theme.background.light};
+    border-color: ${props => props.theme.colors.muted};
+    align-items: center;
+    padding: 0 ${props => props.theme.space[3]};
+`;
+export const ChipDeleteContainer = styled(props => <Button {...props} />)`
+    border-radius: ${props => props.theme.radii.round};
+    height: 20px;
+    width: 20px;
+    padding: 0;
+    margin: 0;
+    align-items: center;
+    justify-content: center;
+    background: lightgrey;
+    border: none;
+`;
+export const ChipTitle = styled(props => <Typography {...props} />)`
+    margin-inline-end: ${props => props.theme.space[4]};
+`;
+
+interface ChipProps {
+    title: string;
+    onDelete: () => void;
+}
+
+const Chip: React.FC<ChipProps> = ({ title, onDelete }) => {
+    return (
+        <ChipContainer>
+            <ChipTitle variant="subtitle">{title}</ChipTitle>
+            <ChipDeleteContainer variant="primary" onClick={onDelete}>
+                <BsX />
+            </ChipDeleteContainer>
+        </ChipContainer>
+    );
+};
+
+export default Chip;
--- a/react/src/components/Chip.tsx
+++ b/react/src/components/Chip.tsx
@@ -9,11 +9,13 @@ export const ChipContainer = styled(props => <Flex {...props} />)`
     border-color: ${props => props.theme.colors.muted};
     align-items: center;
     padding: 0 ${props => props.theme.space[3]};
+    margin: ${props => props.theme.space[2]};
+    height: fit-content;
 `;
 export const ChipDeleteContainer = styled(props => <Button {...props} />)`
     border-radius: ${props => props.theme.radii.round};
-    height: 20px;
-    width: 20px;
+    height: 10px;
+    width: 10px;
     padding: 0;
     margin: 0;
     align-items: center;
@@ -23,6 +25,8 @@ export const ChipDeleteContainer = styled(props => <Button {...props} />)`
 `;
 export const ChipTitle = styled(props => <Typography {...props} />)`
     margin-inline-end: ${props => props.theme.space[4]};
+    margin-top: ${props => props.theme.space[3]};
+    margin-bottom: ${props => props.theme.space[3]};
 `;
 
 interface ChipProps {
--- a/react/src/components/Chip.tsx
+++ b/react/src/components/Chip.tsx
@@ -36,7 +36,7 @@ interface ChipProps {
 
 const Chip: React.FC<ChipProps> = ({ title, onDelete }) => {
     return (
-        <ChipContainer>
+        <ChipContainer justifyContent="space-between" alignItems="center">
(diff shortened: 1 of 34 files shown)


jennyziyi-xu:
Looks great! For `start` and `end` columns, the filter is based on exact matches, which might be the design choice you had. 

However now, when a user starts to type digits into the filter for `start`, all the rows will disappear because the filter is based on exact matches. Since all the rows disappear, the user might have a hard time completing the filter with the rest of the digits. This is just a thought, I am not sure how this would impact an actual user. :)

server/src/models/utils/getCoordinates.ts
```diff
@@ -24,8 +24,8 @@ const getCoordinates = (result: VariantQueryDataResult[]) => {
   });
 
   return {
-    start: start,
-    end: end,
+    start,
+    end,
     coordinates: coordinates,
```
jennyziyi-xu:
```suggestion
    coordinates,
```

react/src/components/Table/Table.tsx
```diff
@@ -512,7 +507,32 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
...
+                                                                    {/* 
+                                                                        Filter icon for filtering individual columns
+                                                                    */}
+                                                                    {!column.disableFilters && (
+                                                                        <FilterPopover
+                                                                            columns={visibleColumns.filter(
+                                                                                c =>
+                                                                                    c.id ===
+                                                                                    column.id
+                                                                            )}
```
jennyziyi-xu: I think filtering through `visibleColumns` is redundant because we check if the column is redundant at line 486. 

Also, `columns` property for this `FilterPopoever` component will always only contain 1 column whereas its type is set to be an array of columns. Maybe you can check if `columns` property can be reduced to 1 column without impacting other components.

jennyziyi-xu (changes requested):

jennyziyi-xu:
Screen Shot 2022-02-16 at 1 46 50 PM

Adding a sorting icon to all the columns is an excellent idea! This would also introduce confusion sometimes. For example, in this screenshot, the table is sorted by `end` but it's impossible to tell by referring to the headers and the icons beside them.

react/src/components/Table/Table.tsx
hannahle: True!

hannahle:
> Screen Shot 2022-02-16 at 1 46 50 PM
> 
> Adding a sorting icon to all the columns is an excellent idea! This would also introduce confusion sometimes. For example, in this screenshot, the table is sorted by `end` but it's impossible to tell by referring to the headers and the icons beside them.

Yeah, I thought about that too. I'll brainstorm another design and push it up.

jennyziyi-xu:
https://user-images.githubusercontent.com/56279639/154574113-11526fba-bd66-4e93-acdf-75689e9f2666.mov

I am not sure if this is the expected behavior. The filter type does not persist. If the filter was originally ">=" or "<=", they become "==" by default.

jennyziyi-xu:
I like the new design for the sorting UI! It can be clearly seen which column is sorted.

react/src/components/Chip.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { BsX } from 'react-icons/bs';
+import styled from 'styled-components/macro';
+import { Button, Flex, Typography } from './index';
+
+export const ChipContainer = styled(props => <Flex {...props} />)`
```
kevinlul, line 6: I _think_ in styled-components, if you just do `styled(Flex)`, the props should get passed through as Flex is used as the component rather than the anonymous one defined here?

react/src/components/Table/FilterPopover.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            animate={{ opacity: 1, scale: 1, x: 0, y: 0 }} // animate to
+                            exit={{ opacity: 0, scale: 0.85, x: 0, y: 20 }} // animate exit
+                            transition={{
+                                type: 'spring',
+                                stiffness: 800,
+                                damping: 35,
+                            }}
+                            style={{
+                                ...style,
+                                ...{
```
kevinlul: don't need to use the spread operator here, can just have the properties in the object literally

react/src/components/Table/Table.tsx
```diff
@@ -556,7 +553,28 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                                                                 <Flex
                                                                     alignItems="center"
                                                                     justifyContent="center"
+                                                                    nowrap
                                                                 >
+                                                                    {/* 
+                                                                        Filter icon for filtering individual columns
+                                                                    */}
```
kevinlul, line 560: This component may be getting too complicated if the indentation is going so far that the formatter is doing this to comments.

react/src/components/Table/TableFilter/InputComparisonDropdown.tsx
```diff
@@ -50,17 +52,23 @@ const COMPARISON_OPTIONS: ComparisonOption[] = [
...
     const [open, setOpen] = useState(false);
-    const [sign, setSign] = useState<keyof ComparisonType>('equal');
+    const [sign, setSign] = useState<keyof ComparisonType>(value);
 
     const ref = React.useRef() as React.MutableRefObject<HTMLDivElement>;
 
     useClickAway(ref, () => setOpen(false));
 
-    // Ensure value returned is a number and not +/-Infinity
+    // Ensure value returned is a number and not +/undefined
```
kevinlul: remove `+/`

react/src/components/Chip.tsx
hannahle: When I changed it to only `styled(Flex)`, I got this error: 
Screen Shot 2022-02-18 at 1 49 36 PM

I think `styled(Flex)` only works when you're not using the props of `Flex`.

react/src/components/Table/Table.tsx
hannahle: Yeah.... I'll make a separate issue for thinking about how to make `Table.tsx` less complicated.

jennyziyi-xu (approved):
LGTM!

PR #197, opened 2022-02-15
React: Collapse duplicate variant rows into one if Case Details section is collapsed. 

jennyziyi-xu:
-  Added a hidden column `unique_id`. Assign `unique_id` to the rows in increasing order according to the order of `chromosome`, `ref`, `alt`, `start`, `end`. The initial/default sorting scheme for the React Table is by ascending order of `unique_id`. 
- Implemented the duplicate variants collapsing feature when the Case Details section is collapsed. Note that the function `prepareData` returns an array of row indices for those rows that will be displayed when the Case Details section is collapsed. Thus, in the rendering, we only need to check if the current row index is in this array returned from function`prepareData`. 

Comments:

-  The if-else conditions in function`prepareData` are not shortened to retain readability, but if needed, the formal expression can be reduced. 
- @hannahle Your recent implementation of result display (ie: "n total variants found") would need to be changed to avoid confusion. 
- Users can hold the "shift" key to sort multiple columns at the same time. I suggest that we add a note on the UI to inform the user about this feature. @hannahle Any thoughts? 
- If other column(s) are sorted, `unique_id` column will not be sorted anymore. However, I think the current UI display when sorting different columns produces the original behavior we want.

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -91,12 +91,13 @@ const prepareData = (queryResult: VariantQueryDataResult[]): [ResultTableColumns
 
     sortedQueryResult.forEach(d => {
         const { ref, alt, start, end } = d.variant;
+
         if (JSON.stringify(currVariant) !== JSON.stringify({ ref, alt, start, end })) {
+            currVariant = { ref, alt, start, end };
             currUniqueId += 1;
             uniqueVariantIndices.push(currRowId);
-            currVariant = { ref, alt, start, end };
         }
-        const id = currUniqueId;
+
         if (d.variant.callsets.length) {
             result.push.apply(
                 result,
@@ -108,13 +109,13 @@ const prepareData = (queryResult: VariantQueryDataResult[]): [ResultTableColumns
                                 ...cs.info,
                                 ...flattenBaseResults(d),
                             },
-                            id
+                            currUniqueId
                         )
                     )
             );
             currRowId += d.variant.callsets.length;
         } else {
-            result.push(addAdditionalFieldsAndFormatNulls(flattenBaseResults(d), id));
+            result.push(addAdditionalFieldsAndFormatNulls(flattenBaseResults(d), currUniqueId));
             currRowId += 1;
         }
     });
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -92,6 +92,7 @@ const prepareData = (queryResult: VariantQueryDataResult[]): ResultTableColumns[
 const Table: React.FC<TableProps> = ({ variantData }) => {
     const [advancedFiltersOpen, setadvancedFiltersOpen] = useState<Boolean>(false);
 
+    // transform data from VariantQueryDataResult to ResultTableColumns format.
     const tableData = useMemo(() => prepareData(variantData), [variantData]);
 
     const sortByArray = useMemo(
@@ -389,23 +390,36 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
         useGlobalFilter,
         useSortBy,
         useExpanded,
-        usePagination
+        usePagination,
+        hooks => {
+            hooks.useInstance.push((Instance) => {
+                const uniqueVariantFields = ["chromosome", "start", "end", "ref", "alt"];
+    
+                let rowSpanHeaders :any = {};  // specify the type here! 
+                uniqueVariantFields.forEach((field) => {
+                    rowSpanHeaders[field] = null;
+                })
+                
+                Object.assign(Instance, {rowSpanHeaders});
+            });
+        }
     );
 
     const {
         getTableProps,
         getTableBodyProps,
-        headerGroups,
-        page,
+        headerGroups,      // useTable 
+        page,               // usePagination 
         state,
         setFilter,
         setAllFilters,
         setGlobalFilter,
         prepareRow,
         preFilteredRows,
-        toggleHideColumn,
-        visibleColumns,
+        toggleHideColumn,       // useTable 
+        visibleColumns,         // useTable
         rows,
+        rowSpanHeaders
     } = tableInstance;
 
     const { filters, globalFilter } = state;
@@ -456,26 +470,28 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                     setFilter={setFilter}
                 />
             )}
-
-            <ScrollContainer ignoreElements="p, th" hideScrollbars={false} vertical={false}>
+            <ScrollContainer ignoreElements="p, th" hideScrollbars={false} vertical={false}>  
                 <Styles>
                     <table {...getTableProps()}>
                         <THead>
+                            {/* Headergroup_0 contains Variant, case details, variant details. Headergroups_1 contains the columns under Variant.   */}
                             {headerGroups.map(headerGroup => {
                                 // https://github.com/tannerlinsley/react-table/discussions/2647
                                 const { key, ...restHeaderGroupProps } =
-                                    headerGroup.getHeaderGroupProps();
+                                    headerGroup.getHeaderGroupProps();                           
                                 return (
                                     <motion.tr layout key={key} {...restHeaderGroupProps}>
+                                        {/* Loop through one header group at one time */}
                                         {headerGroup.headers.map(column => {
+                                            console.log(column);
                                             const { key, ...restHeaderProps } =
                                                 column.getHeaderProps(
                                                     column.getSortByToggleProps({
                                                         title: undefined,
                                                     })
                                                 );
-                                            return (
-                                                <TH key={key} {...restHeaderProps}>
+                                        // This is a function call for each column. 
+                                            return (<TH key={key} {...restHeaderProps}>
                                                     <AnimatePresence initial={false}>
                                                         {column.isVisible && (
                                                             <motion.section
@@ -526,7 +542,6 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                                                                                 : ''
                                                                         }`}
                                                                     />
-
                                                                     {isHeader(column) &&
                                                                         (isHeaderExpanded(
                                                                             column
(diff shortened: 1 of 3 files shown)


react/src/components/Table/Table.tsx
```diff
@@ -59,6 +59,7 @@ export interface ResultTableColumns extends FlattenedQueryResponse {
     aaChange: string;
     emptyCaseDetails: string;
     emptyVariationDetails: string;
+    unique_id: number;
```
hannahle: Can we make this camel case? (uniqueId)

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
-                        ...cs.info,
-                        ...flattenBaseResults(d),
-                    })
-                );
+/* 1, Sort queryResult in ascending order according to variant's chromosome, ref, alt, start,end. 
+2, flatten data and compute values as needed 
+(note that column display formatting function should not alter values for ease of export) */
+const prepareData = (queryResult: VariantQueryDataResult[]): [ResultTableColumns[], number[]] => {
+    const sortedQueryResult = [...queryResult].sort(function (a, b) {
+        if (a.variant.referenceName < b.variant.referenceName) {
```
hannahle: Can we move this `sort` functionality to `tableHelpers.ts`? Ideally, we want Table.tsx to focus on styling code and React logic and abstract the logic away to another file.

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
+            return -1;
+        } else if (
+            a.variant.referenceName === b.variant.referenceName &&
```
hannahle: We don't need a check for `referenceName` because user is only allowed to search for one gene at the time. Reference name refers to the chromosome that gene is on, so all of our variants would have the same referenceName.

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
+    var uniqueVariantIndices: Array<number> = []; // contains indices of first encountered rows that represent unique variants.
+    const n = sortedQueryResult.length;
+    var currVariant = { chromosome: 'temp', ref: 'temp', alt: 'temp', start: 0, end: 0 };
+    var currUniqueId = 0;
+    var currRowId = 0;
+
+    for (let i = 0; i < n; i++) {
+        const d = sortedQueryResult[i];
+        if (
+            d.variant.referenceName !== currVariant['chromosome'] ||
```
hannahle: Same comment as above applies here. We can remove all the checks for referenceName/chromosome.

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
+            d.variant.ref !== currVariant['ref'] ||
+            d.variant.alt !== currVariant['alt'] ||
+            d.variant.start !== currVariant['start'] ||
+            d.variant.end !== currVariant['end']
+        ) {
+            currUniqueId += 1;
+            uniqueVariantIndices.push(currRowId);
+            currVariant['chromosome'] = d.variant.referenceName;
```
hannahle: Can you explain why we would need to assign values to `currVariant` here, since it's not used anywhere else after this code block?

hannahle: Actually, I got it. You are trying to reuse the value for comparison in the `if` condition above to determine if a variant is unique.

jennyziyi-xu: yes, that's right.

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
         } else {
-            return addAdditionalFieldsAndFormatNulls(flattenBaseResults(d));
+            return 1;
         }
     });
+
+    var result: Array<ResultTableColumns> = [];
+    var uniqueVariantIndices: Array<number> = []; // contains indices of first encountered rows that represent unique variants.
+    const n = sortedQueryResult.length;
+    var currVariant = { chromosome: 'temp', ref: 'temp', alt: 'temp', start: 0, end: 0 };
```
hannahle:
```suggestion
   type Variant = Pick<VariantResponseFields, 'referenceName' | 'alt' | 'ref' | 'start' | 'end'>;
    var currVariant: Variant  = {} as Variant;
```

Note: You should move the `type Variant` to top of file where all the other types are declared.

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
+            d.variant.referenceName !== currVariant['chromosome'] ||
+            d.variant.ref !== currVariant['ref'] ||
+            d.variant.alt !== currVariant['alt'] ||
+            d.variant.start !== currVariant['start'] ||
+            d.variant.end !== currVariant['end']
+        ) {
+            currUniqueId += 1;
+            uniqueVariantIndices.push(currRowId);
+            currVariant['chromosome'] = d.variant.referenceName;
+            currVariant['ref'] = d.variant.ref;
```
hannahle: With the suggestion I made above, this can be rewritten in a shorter way:

```
const { referenceName, ref, alt, start, end } = d.variant;
currVariant = { referenceName, ref, alt, start, end }
```

react/src/components/Table/Table.tsx
```diff
@@ -71,34 +72,103 @@ const resolveSex = (sexPhenotype: string) => {
...
+            a.variant.alt === b.variant.alt &&
+            a.variant.start < b.variant.start
+        ) {
+            return -1;
+        } else if (
+            a.variant.referenceName === b.variant.referenceName &&
+            a.variant.ref === b.variant.ref &&
+            a.variant.alt === b.variant.alt &&
+            a.variant.start === b.variant.start &&
+            a.variant.end < b.variant.end
```
hannahle: We can combine all the `else if` clause to: 

```
if (
            a.variant.alt <= b.variant.alt ||
            a.variant.ref <= b.variant.ref ||
            a.variant.start <= b.variant.start ||
            a.variant.end <= b.variant.end
        ) {
            return -1;
        } else {
            return 1;
        }
```

react/src/components/Table/Table.tsx
```diff
@@ -592,6 +667,18 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                                 page.map(row => {
                                     prepareRow(row);
                                     const { key, ...restRowProps } = row.getRowProps();
+                                    const caseDetailsCol = headerGroups[0].headers.find(
+                                        header => header.Header === 'Case Details'
+                                    );
+                                    // Display only one row per variant if Case Details Section is collapsed.
+                                    if (caseDetailsCol && !isHeaderExpanded(caseDetailsCol)) {
+                                        if (
+                                            uniqueVariantIndices.find(i => i === row?.index) ===
```
hannahle: In our codebase, we often write `!uniqueVariantIndices.find(i => i === row?.index)` which is the equivalent of `uniqueVariantIndices.find(i => i === row?.index) === undefined`

react/src/components/Table/Table.tsx
```diff
@@ -592,6 +667,18 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                                 page.map(row => {
                                     prepareRow(row);
                                     const { key, ...restRowProps } = row.getRowProps();
+                                    const caseDetailsCol = headerGroups[0].headers.find(
```
hannahle: Can you make a function called `isCaseDetails` to wrap this logic and move it to the `tableHelpers.ts` file?

react/src/utils/tableHelpers.ts
```diff
@@ -54,6 +56,7 @@ export const addAdditionalFieldsAndFormatNulls = (
         ...reformatted,
         emptyCaseDetails: '',
         emptyVariationDetails: '',
+        unique_id: currId,
```
hannahle: If you call the parameter you're passing into the function `unique_id` instead of `currId`, you would only need to write `unique_id` in this line versus `unique_id: currId` because JSON understands to take the value of the property with the same name.

react/src/components/Table/Table.tsx
jennyziyi-xu: Yes, that's a good point, will do. Since this `sort` functionality is only used in this one place, I didn't move it to `tableHelpers.ts`.

react/src/components/Table/Table.tsx
jennyziyi-xu: Yes, that makes sense.

hannahle:
_To respond to your comments:_

> * The if-else conditions in function`prepareData` are not shortened to retain readability, but if needed, the formal expression can be reduced.

- I made more detailed suggested changes on my comment.

> * @hannahle Your recent implementation of result display (ie: "n total variants found") would need to be changed to avoid confusion.

- Will do! I asked in the G4RD group chat what the Matt and Magda want displayed here.

> * Users can hold the "shift" key to sort multiple columns at the same time. I suggest that we add a note on the UI to inform the user about this feature. @hannahle Any thoughts?

- I will have to double check the behaviour on my current branch (PR #196 ) for this. It's out of scope for your PR but I will have to look to see how it can be incorporated if ever to my PR. 

> * If other column(s) are sorted, `unique_id` column will not be sorted anymore. However, I think the current UI display when sorting different columns produces the original behavior we want.

- This is okay! I don't think the users would expect the variants to be grouped anymore now that they're trying to sort columns in case details.

react/src/components/Table/Table.tsx
jennyziyi-xu: This is not equivalent to the original if-else block. You could verify that with your suggestion, the sorting order is not what we would expect.

hannahle: Ahah - you're right. I wonder if there's another way we can condense it.

jennyziyi-xu: As I've mentioned in this PR, I can always use formal logic to reduce the original if-else block, but it will not really be readable. Without extensive comments, future developers would have a hard time understanding what the code is doing.

hannahle: Have you looked at this? https://dev.to/markbdsouza/js-sort-an-array-of-objects-on-multiple-columns-keys-2bj1

react/src/components/Table/Table.tsx
jennyziyi-xu: `!uniqueVariantIndices.find(i => i === row?.index)` would ignore rows that have `index=0`. 

`index` is a property for rows: [link]( https://react-table.tanstack.com/docs/api/useTable#row-properties). With the underlying implementation, the index starts with 0. It might be more robust to conserve the default implementation.

react/src/components/Table/Table.tsx
jennyziyi-xu: Just to make sure, do we want to sort `ref` or `alt` first?

react/src/components/Table/Table.tsx
jennyziyi-xu: Ok, got it. Thanks!

react/src/components/Table/Table.tsx
hannahle: Either is fine!

react/src/components/Table/Table.tsx
hannahle: Ah that's right. Agreed.

react/src/components/Table/Table.tsx
jennyziyi-xu: > Have you looked at this? https://dev.to/markbdsouza/js-sort-an-array-of-objects-on-multiple-columns-keys-2bj1

Yes, this is a better way to condense the experession. Thanks very much!

PR #198, opened 2022-02-17
React: Add zygosity count columns, alternate rows' background colour 

jennyziyi-xu:
1. Added two columns to Variant Details group - `heterozygous count` and `homozygous count`. Each column represents how many heterozygous traits there are for a given variant.
2. Add alternating colors for the rows. Behavior under different scenarios: 
- When CaseDetails is not expanded, adjacent rows have different colors. 
- When the user expands case details, the same variant has the same color. 
- When the user clicks on a column to sort inside Case Details, now not all variants would be in the same group anymore. But the ones that do coincidentally group together would still have the same color.

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -28,13 +28,11 @@ import {
     VariantResponseInfoFields,
 } from '../../types';
 import {
-    addAdditionalFieldsAndFormatNulls,
     calculateColumnWidth,
-    flattenBaseResults,
     isCaseDetailsCollapsed,
     isHeader,
     isHeaderExpanded,
-    sortQueryResult,
+    prepareData,
 } from '../../utils';
 import { Button, Flex, InlineFlex, Tooltip, Typography } from '../index';
 import { Column } from '../Layout';
@@ -61,10 +59,10 @@ export interface ResultTableColumns extends FlattenedQueryResponse {
     aaChange: string;
     emptyCaseDetails: string;
     emptyVariationDetails: string;
+    homozygousCount?: number;
+    heterozygousCount?: number;
     uniqueId: number;
 }
-
-type Variant = Pick<VariantResponseFields, 'ref' | 'alt' | 'start' | 'end'>;
 
 const resolveSex = (sexPhenotype: string) => {
     if (sexPhenotype.toLowerCase().startsWith('m') || sexPhenotype === 'NCIT:C46112') {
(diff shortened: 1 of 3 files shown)


react/src/utils/tableHelpers.ts
```diff
@@ -108,3 +111,91 @@ export const sortQueryResult = (queryResult: VariantQueryDataResult[]) => {
     );
     return sortedQueryResult;
 };
+
+export const isHeterozygous = (zygosity: String | null | undefined) => {
+    if (zygosity?.toLowerCase().includes('het')) {
+        return true;
+    }
+    return false;
```
kevinlul: This and the following line can be just one-line if you do `return !!zygosity?.toLowerCase().includes('het')` to get rid of the nullish values.

react/src/utils/tableHelpers.ts
```diff
@@ -108,3 +111,91 @@ export const sortQueryResult = (queryResult: VariantQueryDataResult[]) => {
     );
     return sortedQueryResult;
 };
+
+export const isHeterozygous = (zygosity: String | null | undefined) => {
```
kevinlul: You should use `string` as the typing, not the `String` object.

react/src/utils/tableHelpers.ts
```diff
@@ -108,3 +111,91 @@ export const sortQueryResult = (queryResult: VariantQueryDataResult[]) => {
...
+    var currVariant = {} as Variant;
+    var currUniqueId = 0;
+    var currRowId = 0;
+    var currHomozygousCount = 0;
+    var currHeterozygousCount = 0;
+
+    sortedQueryResult.forEach(d => {
+        const { ref, alt, start, end } = d.variant;
+
+        if (JSON.stringify(currVariant) !== JSON.stringify({ ref, alt, start, end })) {
```
kevinlul: This could be an expensive comparison to do in a loop, are you sure about this?

react/src/components/Table/Table.tsx
```diff
@@ -364,6 +317,18 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         width: 79,
                     },
                     { accessor: 'transcript', id: 'transcript', Header: 'transcript', width: 150 },
+                    {
+                        accessor: 'homozygousCount',
+                        id: 'homozygousCount',
+                        Header: 'homozygous count',
```
hannahle: Capitalize: Homozygous Count

react/src/components/Table/Table.tsx
```diff
@@ -364,6 +317,18 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
...
+                        width: 150,
+                    },
+                    {
+                        accessor: 'heterozygousCount',
+                        id: 'heterozygousCount',
+                        Header: 'heterozygous count',
+                        width: 150,
```
hannahle: You might want to use `getColumnWidth` here. The function calculates the content to adjust width accordingly, versus hardcoding the width.

react/src/components/Table/Table.tsx
jennyziyi-xu: Just to confirm, are we sure about this? All the other headers are not capitalized.

hannahle: Yup! I think the other ones are not capitalized (like `gnomadHom`) because it's the standard way the gnomAD database writes it

react/src/utils/tableHelpers.ts
jennyziyi-xu: Are you suggesting that using `JSON.stringify` to compare objects is expensive?

kevinlul: Yes, because it is being done in a loop. You have the overhead of the conversion rather than comparing four properties in memory. Furthermore (unrelated to performance), this relies on initializing `currVariant` below in the same order as the construction above.

jennyziyi-xu: Ok got it. I've changed it to direct comparisons between fields.

hannahle (approved):
LGTM!

PR #202, opened 2022-02-25
React: Upgrade CRA to 5.0.0

jennyziyi-xu:
- Upgraded `react-scripts` to version 5.0.0.
- `WBPACK_DEFAULT_EXPORT` error is introduced. Source of the error: In `App.tsx`: `import {AboutPage, VariantQueryPage} from '.'`. In `index.ts`: `import AppPage from './App'`. App.tsx imports from index.ts, index.ts imports from App.tsx, which is a cyclical dependency. Thanks to @kevinlul and @hannahle for catching this error.  
- To fix the error, in  `App.tsx`, make direct imports instead of importing from `index.ts`.


kevinlul (commented):
I believe you may need to update some other packages in lockstep and also get rid of the forced resolution for babel-loader, not sure why it's there

jennyziyi-xu:
The warnings for React missing keys are bugs from the dependency `eslint-plugin-react`. This bug was fixed last Friday likely after I signed off, which explains why the warnings are gone today whereas I didn't change anything in comparison to last Friday. Bug fix: https://github.com/yannickcr/eslint-plugin-react/commit/13d8df686945b64a727f378c27305e1487af30fe

hannahle (approved):
LGTM!

PR #211, opened 2022-03-10
React/Node: Resolve gene name

hannahle:
From discussion with our collaborators, we came to the conclusion that Ensembl ID is not a reliable point of reference for the identity of an actual gene, and should not be used in our search query to the `remote-test` and future `g4rd` node. This is due to ensembl IDs changing over time, and users not paying attention to which ensembl version they're using. 

To improve user's interpretability of our gene search, the following changes have been implemented: 
- Helper message informing users multiple aliases are available
- Specified chromosome, start, and end position since one gene name can refer to completely different genes on different chromosomes.
- Removed Ensembl ID from UI.
- Removed Ensembl ID as a search param to `remote-test` endpoint.
- Modified `remote-test` endpoint to only take in gene name and assembly.
- Added check for whether a gene is canonical. To elaborate, a gene such as `ABHD16A` refers to 6 chromosomes, but 5 of them is in regions that are major histocompatibility. These regions are highly polymorphic (highly variable between individuals), not representative of many individuals, and therefore should be removed from search.

### DIFF
--- a/react/src/components/GeneSearch.tsx
+++ b/react/src/components/GeneSearch.tsx
@@ -4,7 +4,7 @@ import { useFetchAutocompleteQuery } from '../apollo/hooks';
 import { AssemblyId } from '../types';
 import ComboBox from './ComboBox';
 import { SelectableListItem } from './SelectableList';
-
+import { Background, Typography } from '../components';
 interface HitPosition {
     chr: string;
     start: number;
@@ -76,9 +76,9 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
                                 position: `${e.chr}:${e.start}-${e.end}`,
                             },
                             id: i + eid,
-                            label: Array.isArray(hit.ensembl)
-                                ? `${hit.symbol.toUpperCase()} - ${genes.ensembl[eid].gene}`
-                                : hit.symbol.toUpperCase(),
+                            label: `${hit.symbol.toUpperCase()} (Chromosome: ${e.chr}, Start: ${
+                                e.start
+                            }, End: ${e.end})`,
                         }));
                 })
                 .flat(),
@@ -103,15 +103,29 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
     }, [geneName, assembly, autocompleteResults, formatAutocompleteOptions]);
 
     return (
-        <ComboBox
-            options={options}
-            loading={autocompleteLoading}
-            onChange={term => onChange(term)}
-            onSelect={(item: SelectionValue) => onSelect(item)}
-            placeholder="Gene Search"
-            searchable
-            value={geneName || ''}
-        />
+        <>
+            {options.length > 1 && (
+                <Background
+                    variant="success"
+                    style={{
+                        padding: '0rem 0.75rem',
+                    }}
+                >
+                    <Typography variant="subtitle" success bold>
+                        Multiple gene aliases are found. Please select the appropriate gene.
+                    </Typography>
+                </Background>
+            )}
+            <ComboBox
+                options={options}
+                loading={autocompleteLoading}
+                onChange={term => onChange(term)}
+                onSelect={(item: SelectionValue) => onSelect(item)}
+                placeholder="Gene Search"
+                searchable
+                value={geneName || ''}
+            />
+        </>
     );
 };
 
--- a/react/src/components/GeneSearch.tsx
+++ b/react/src/components/GeneSearch.tsx
@@ -37,11 +37,16 @@ interface GeneSearchProps {
     onChange: (geneName: string) => void;
 }
 
+const isCanonicalRegion = (chr: string) =>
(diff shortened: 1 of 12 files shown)


jennyziyi-xu (commented):
In `GeneSearch.tsx`, should we preserve the interface `Gene`? This interface is just a wrapper around `string` type. Also, this is a minor issue but using `Gene` to denote the type for Genome Assembly can be confusing.

react/src/components/GeneSearch.tsx
```diff
@@ -24,7 +23,7 @@ interface AutocompleteResults {
         }[];
     };
 }
-interface SelectionValue {
+export interface GeneSelectionValue {
     ensemblId?: string;
```
jennyziyi-xu:
```suggestion
```

react/src/components/GeneSearch.tsx
```diff
@@ -64,17 +66,21 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
...
-                            ? `${hit.symbol.toUpperCase()} - ${genes.ensembl[eid].gene}`
-                            : hit.symbol.toUpperCase(),
-                    }));
+                    return (is38 ? genomic_pos : genomic_pos_hg19)
+                        .filter(g => isCanonicalRegion(g.chr))
+                        .map((e, eid) => {
+                            return {
+                                value: {
+                                    name: genes.symbol.toUpperCase(),
+                                    ensemblId: '',
```
jennyziyi-xu:
```suggestion
```

jennyziyi-xu (changes requested):

jennyziyi-xu:
Suggestion: delete `ensemblId` in `ssmp-schema.yml`, `server/src/type.ts` and `server/src/typeDefs.ts`.

test-node/src/server.ts
```diff
@@ -55,7 +55,6 @@ app.get(
 
     const result = await getStagerData(
       geneName as string,
-      ensemblId as string,
       assemblyId as string
```
jennyziyi-xu:
```suggestion
```

test-node/src/server.ts
```diff
@@ -68,7 +67,7 @@ app.get(
   }
 );
 
-const getStagerData = async (geneName: string, ensemblId: string, assemblyId: string) => {
+const getStagerData = async (geneName: string, assemblyId: string) => {
```
jennyziyi-xu, line 65:
```suggestion
const getStagerData = async (geneName: string) => {
```

server/src/resolvers/getVariantsResolver/adapters/remoteTestNodeAdapter.ts
```diff
@@ -113,9 +113,8 @@ const getRemoteTestNodeQuery = async (args: QueryInput): Promise<VariantQueryRes
   let remoteTestNodeQueryError: RemoteTestNodeQueryError | null = null;
 
   try {
-    console.log(process.env.TEST_NODE_URL);
     remoteTestNodeQueryResponse = await axios.get<StagerVariantQueryPayload[]>(
```
jennyziyi-xu, line 116: `ensembl_id` is still a field in `StagerVariantQueryPayload`, maybe consider deleting it.

react/src/pages/VariantQueryPage.tsx
```diff
@@ -188,46 +175,32 @@ const VariantQueryPage: React.FC<{}> = () => {
...
                             geneName={queryOptionsForm.gene.value}
-                            onChange={geneName =>
-                                updateQueryOptionsForm({ gene: geneName, ensemblId: '' })
-                            }
+                            onChange={geneName => updateQueryOptionsForm({ gene: geneName })}
                             onSelect={val => {
-                                const { position, ensemblId } = val;
+                                const { position } = val;
                                 updateQueryOptionsForm({
                                     gene: val.name,
```
jennyziyi-xu, line 197:
```suggestion
                                const { position, name } = val;
                                updateQueryOptionsForm({
                                    gene: name,
```

react/src/components/GeneSearch.tsx
```diff
@@ -98,15 +104,17 @@ const GeneSearch: React.FC<GeneSearchProps> = ({ assembly, geneName, onChange, o
...
-        <ComboBox
-            options={options}
-            loading={autocompleteLoading}
-            onChange={term => onChange(term)}
-            onSelect={(item: SelectionValue) => onSelect(item)}
-            placeholder="Gene Search"
-            searchable
-            value={geneName || ''}
-        />
+        <>
```
jennyziyi-xu:
```suggestion
```

I don't think it's necessary to wrap `ComboBox.`

jennyziyi-xu (changes requested):

server/src/resolvers/getVariantsResolver/adapters/remoteTestNodeAdapter.ts
hannahle: Yes, we can delete it. I've switched Stager search to purely gene-based.

jennyziyi-xu (approved):
LGTM!

PR #221, opened 2022-03-30
React: Make columns re-orderable

### LINKED ISSUE #213: React: Make columns re-orderable

jennyziyi-xu:
Columns re-ordering logic is added in `ColumnVisibilityModal`. The columns in the table are re-ordered only when the user clicks `Save` in the modal. 
Closes #213


kevinlul (commented):
If the only purpose of the modal is to reorder columns, maybe title it that way and label the primary action button "reorder" or "apply"?

jennyziyi-xu:
Screen Shot 2022-03-30 at 4 29 07 PM


This modal also allows a user to toggle the visibility of a column.

jennyziyi-xu:
After merging this PR in, the logic for toggling column visibility will be addressed and modified in #222

react/src/components/Table/ColumnVisibilityModal.tsx
```diff
@@ -1,70 +1,152 @@
...
 export default function ColumnVisibilityModal<T extends {}>({
     headerGroups,
     toggleGroupVisibility,
     toggleHideColumn,
+    allColumns,
+    setColumnOrder,
 }: ColumnVisibilityModalProps<T>) {
     const [showModal, setShowModal] = useState<boolean>(false);
+    const [order, setOrder] = useState<ColumnInstance<T>[]>([]);
+    const reorder = (startIndex: number, endIndex: number): ColumnInstance<T>[] => {
```
hannahle: Maybe name this `prevPos`, `newPos` for readability? I was a bit confused by `startIndex`, `endIndex`

hannahle:
A couple small UI things: 
- When users drag the header to another column group, we should show an error message saying something like "Can't drag your column to another group". As the user drags the header above the ineligible column group before "dropping" them, the background of that column group should be red.
- Instead of "Reorder", we can have "Apply" instead because there are multiple functionalities?
- It's a bit confusing that the invisible columns don't have drag. I think we should have the drag for visible columns purple, and still display the drag for invisible columns but have them grey instead. When user hovers above the grey drag icon, it has a hover message that says "Only visible columns are draggable". That way users know they have to turn on the columns first before dragging them.

react/src/components/Table/ColumnVisibilityModal.tsx
```diff
@@ -1,70 +1,152 @@
...
             <Modal
                 active={showModal}
                 hideModal={() => setShowModal(false)}
                 title="Customize Columns"
+                footer="Apply"
+                onClick={() => {
+                    setColumnOrder(order.map(o => o.id));
+                    setShowModal(false);
+                }}
+                helperText="Only columns within the same group can be reordered."
```
hannahle: Because the "Customize Columns" module is for both toggling visibility and moving columns around, can we adjust the helperText to reflect that? And we can then add "note that only columns within the same group can be reordered"

react/src/components/Table/ColumnVisibilityModal.tsx
```diff
@@ -1,70 +1,152 @@
...
+                                    {...droppableProvided.droppableProps}
+                                >
+                                    <Checkbox
+                                        label={g.Header as string}
+                                        checked={g.isVisible}
+                                        onClick={() => toggleGroupVisibility(g)}
+                                    />
+                                    {order.map(
+                                        (c, id) =>
+                                            g.columns?.find(column => column.id === c.id) &&
```
hannahle, line 80: The code here is getting pretty hard to read. Note to self to open another issue to refactor react-table

jennyziyi-xu: Sure, you can also add on to this issue opened earlier: [https://github.com/ccmbioinfo/osmp/issues/216](https://github.com/ccmbioinfo/osmp/issues/216)

jennyziyi-xu:
Screen Shot 2022-04-05 at 9 59 32 AM

A tooltip is added to inform the user about the functionalities of the modal. We also need to add drag handles for columns that are invisible and implement the functionalities accordingly. I will do these after refactoring `Customize Columns` modal to optimize the response time when the user toggles the visibility of a column in the modal.

hannahle:
LGTM!

hannahle (approved):

PR #225, opened 2022-04-04
React: Shrink columns width conditionally 

### LINKED ISSUE #214: React: Shrink columns' width.
- [x] Shrink width of individual columns to width of text
- [x] Change 3 columns' name to shrink size:
`Chromosome` -> `chr`
`Homozygous count` -> `Homo Count`
`Heterozygous count` -> `Het Count`

jennyziyi-xu:
- Compute some columns' width based on the length of the column header and the length of the longest data entry for that column. Refactor `CalculateColumnWidth` because the original code that computes the longest data entry for a specific column does not work as expected. 
- In this PR, if there are more than 12 visible columns, the columns have a width equal to what is defined in `const columns = useMemo()`. If there are less than 12 visible columns, the columns will adjust their width to fill 100% of the table. 

closes #214


react/src/components/Table/Table.tsx
```diff
@@ -224,13 +239,19 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         accessor: 'individualId',
                         id: 'individualId',
                         Header: 'Individual ID',
-                        width: getColumnWidth(tableData, 'individualId', 'Individual ID'),
+                        width: getColumnWidth(
+                            Math.max(...tableData.map(row => (row.individualId || '').length)),
```
hannahle: If we are calling `Math.max` for every time we call `getColumnWidth`, can this also be part of the `getColumnWidth` function?

react/src/components/Table/Table.styles.tsx
```diff
@@ -24,14 +24,18 @@ export const TableFilters = styled(props => <Flex {...props} />)`
...
+
+export const Styles = styled.div<StylesProps>`
     padding: 1rem;
     display: block;
     max-width: 100%;
     max-height: 80vh;
 
     table {
-        width: 100%;
+        width: ${props => (props.disableFullWidth ? null : '100%')};
```
hannahle, line 38:
```suggestion
        width: ${props => (!props.disableFullWidth && '100%')};
```

hannahle:
We should modify the spacing for the headers because some part of some columns are hidden. 
Screen Shot 2022-04-04 at 6 32 52 PM

Ideally, the width should fit the icons as well.

react/src/components/Table/Table.tsx
jennyziyi-xu: Yes for sure, this can also be a part of `getColumnWidth` function. I didn't include it in `getColumnWidth` function because we would pass a large size parameter to `getColumnWidth`.

react/src/components/Table/Table.styles.tsx
jennyziyi-xu: That sounds reasonable. However, making this change would result in unexpected behaviors because these two lines of code are not equivalent.

kevinlul (commented):
Maybe don't commit commented code to the mainline; that's why we use Git

jennyziyi-xu:
> Maybe don't commit commented code to the mainline; that's why we use Git

Sorry about that. Some codes were commented out to experiment with a different solution.

hannahle (approved):

PR #227, opened 2022-04-06
React: Add drag handle to invisible columns in Customize Columns modal

### LINKED ISSUE #224: React: Add drag handle to invisible columns in Customize Columns modal
In the Customize Columns modal, make the drag handle purple for visible columns, and make the drag handle grey for invisible columns. 

When a user hovers above the grey drag icon, show a hover message "Only visible columns are draggable". That way users know they have to turn on the columns first before dragging them. Thanks for the suggestions, Hannah.

jennyziyi-xu:
Closes #224 
Screen Shot 2022-04-06 at 4 06 17 PM

In the Customize Columns modal, make the drag handle purple for visible columns, and make the drag handle grey for invisible columns. It is difficult to distinguish different colors so the different drag handle icons are used for visible and invisible columns.

When a user hovers above the grey drag icon, show a hover message "Only visible columns are draggable". That way users know they have to turn on the columns first before dragging them.

### DIFF
--- a/react/src/components/DragHandle.tsx
+++ b/react/src/components/DragHandle.tsx
@@ -19,22 +19,22 @@ interface DragHandleProps {
 }
 
 const DragHandle: React.FC<DragHandleProps> = ({ isVisible, dragHandleProps }) => {
-    return (
-        <>
-            {isVisible && (
+    if (isVisible)
+        return (
+            <div style={{ paddingRight: 40 }}>
                 <Component {...dragHandleProps} isVisible={isVisible}>
                     <RiDragMove2Fill />
                 </Component>
-            )}
-
-            {!isVisible && (
-                <Tooltip helperText="Only visible columns are draggable.">
-                    <Component {...dragHandleProps} isVisible={isVisible}>
-                        <MdDragHandle />
-                    </Component>
-                </Tooltip>
-            )}
-        </>
+            </div>
+        );
+    return (
+        <div style={{ paddingRight: 40 }}>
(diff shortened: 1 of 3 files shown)


react/src/components/DragHandle.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+            {!isVisible && (
+                <Tooltip helperText="Only visible columns are draggable.">
+                    <Component {...dragHandleProps} isVisible={isVisible}>
+                        <MdDragHandle />
+                    </Component>
+                </Tooltip>
+            )}
+        </>
+    );
```
kevinlul: May be more readable to have an if/else on `isVisible`

react/src/components/Table/ColumnVisibilityModal.tsx
```diff
@@ -125,18 +124,14 @@ export default function ColumnVisibilityModal<T extends {}>({
...
-                                                                    style={{
-                                                                        justifyContent: 'flex-end',
-                                                                        display: 'flex',
-                                                                        paddingRight: '40px',
-                                                                    }}
-                                                                >
-                                                                    <MdDragHandle />
-                                                                </div>
-                                                            )}
+                                                            <div style={{ paddingRight: 40 }}>
```
kevinlul: Should the div be part of the DragHandle component?

kevinlul: Should the div be part of the DragHandle component?

PR #228, opened 2022-04-13
Fix Docker

kevinlul:
server:
- copy over dependency fixes
- remove incorrect licence
- set main to actual production image entrypoint file
- remove unused package scripts
- NODE_ENV=local -> NODE_ENV=development for Express
- rewrite Dockerfile
- move INDEX_38 to MinIO

test-node:
- copy over dependency fixes
- fix package name
- set main to actual production image entrypoint file
- remove unused package scripts
- remove graphql reference in copied eslintrc
- create Dockerfile

workflow:
- also build, lint, check formatting for test-node
- allow lint and formatting jobs to fail if they have never been run before
- fix bind mount paths

local compose:
- switch to more standard /usr/src/app mount paths
- use node:16-slim
- use init if running yarn script
- don't unnecessarily use entrypoint
- automatically interpolate MONGO_CONNECTION_STRING
- remove extra bridge network definition
- update Keycloak

staging compose:
- consistent restart and logging configuration
- use :latest tag instead of overwriting :dev tag
- use test-node image, don't mount source code
- update keycloak
- PROJECT_ROOT to help workflow fix bind mount paths


kevinlul:
Stuff I can also help fix
- [x] tsconfig.json
- [ ] unnecessary secrets

server/Dockerfile
```diff
@@ -1,24 +1,30 @@
...
-WORKDIR /usr/app
-
-COPY --from=BUILD_IMAGE /home/node/cadd_wgs_ghr38_index.gz.tbi /home/node/cadd_wgs_ghr38_index.gz.tbi
-COPY --from=BUILD_IMAGE /usr/app/dist ./dist
-COPY --from=BUILD_IMAGE /usr/app/node_modules ./node_modules
-COPY --from=BUILD_IMAGE /usr/app/package.json ./package.json
+# This stage will probably be removed when we index GRCh38 ourselves
+FROM curlimages/curl AS tbi
+WORKDIR /var/tmp
+RUN curl --output cadd_wgs_ghr38_index.gz.tbi https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/whole_genome_SNVs_inclAnno.tsv.gz.tbi
```
hannahle: We can remove the `tbi` step altogether and paste `https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/whole_genome_SNVs_inclAnno.tsv.gz.tbi` as the index file in `fetchCaddAnnotations`, in a similar fashion of how the remote 37 file is passed in.

server/Dockerfile
```diff
@@ -1,24 +1,30 @@
...
+
+# This stage is tagged for local development to guarantee the index is in the container
+FROM node:16-slim AS dev
+WORKDIR /usr/src/app
+COPY --from=tbi --chown=node:node /var/tmp/cadd_wgs_ghr38_index.gz.tbi /home/node
```
hannahle: Can be removed

server/src/resolvers/getVariantsResolver/utils/fetchCaddAnnotations.ts
```diff
@@ -13,7 +13,7 @@ const ANNOTATION_URL_37 =
 
 const INDEX_37_PATH =
   'https://minio.genomics4rd.ca/www-ssmp-dev/whole_genome_SNVs_inclAnno_GRCh37.tsv.gz.csi';
-const INDEX_38_PATH = '/home/node/cadd_wgs_ghr38_index.gz.tbi';
+const INDEX_38_PATH = 'https://minio.genomics4rd.ca/www-ssmp-dev/cadd_wgs_ghr38_index.gz.tbi';
```
hannahle:
```suggestion
const INDEX_38_PATH = 'https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/whole_genome_SNVs_inclAnno.tsv.gz.tbi';
```

kevinlul: Then I remove the index from the frontend static files?

hannahle: Oh, I didn't realize you uploaded the index to the frontend static files?

When I pasted https://minio.genomics4rd.ca/www-ssmp-dev/cadd_wgs_ghr38_index.gz.tbi, got this:

Screen Shot 2022-04-18 at 2 43 55 PM

hannahle: No strong preference on my end- just wanna make sure the file is actually there is all!

kevinlul: It will be deployed with the frontend when this PR is merged like the `.csi` file. As for which server to use, did you confirm if the hosting lab intended for its downloads to be used as a CDN?

hannahle: Can you elaborate on the CDN point?

kevinlul: Does the lab consider directly using the hosted files on its servers by applications acceptable use? Servers and bandwidth don't come for free, so if it's just a download site then it shouldn't be a dependency of this application.

hannahle: I think the collaborators would bias towards using as little resources on the server as possible. The index files are also relatively small (5MB), so I'm not sure how big of a load that is

hannahle: So in conclusion, we should use the remote link that's not in Minio?

hannahle (approved):
LGTM!

PR #234, opened 2022-04-28
Add liftover, Cadd annotations available, Gnomad annotations unavailable. 

jennyziyi-xu:
- Assume that we are querying from remote-test node, perform liftover from assembly GRCh37 to GRCh38. Since the G4RD node is not availble for query, we make this assumption. This is only a temporary working version for demo purpose. 
- Conversion between 1-based position format and 0-based half-open BED format is performed because liftover takes a BED file as input and outputs files in BED format. The conversion assumes that we are dealing with SNV. If we are dealing with multiple nucleotide variants, adjustment needs to be made because the BED format is half open. 
- Cadd annotations are performed, gnomad annotations are not available in the stack yet after discussing with Hannah and will be added later. 
- liftOver and liftover chain files are not added to the production image yet. 

Future steps:

- [ ] Refactor functions and code.
- [ ] Make the liftover step generic, not just for the remote test node. 
- [ ] Generate annotation files for Gnomad annotations.


jennyziyi-xu:
See #244 for a final version of liftover implementation.

PR #236, opened 2022-05-02
Add liftOver executable and chain files to Docker 

jennyziyi-xu:
Add liftOver executable and chain files to local development and production deployment.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -7,11 +7,10 @@ RUN curl -f -L -O https://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/liftOve
     gunzip ./hg19ToHg38.over.chain.gz && \
     chmod +x ./liftOver
 
-# Development image 
+# Development image. Note that libcurl4 is a dependency of liftOver.
 FROM node:16-slim AS dev
-WORKDIR /usr/src/app
-COPY --from=download /var/tmp/liftOver /usr/local/bin 
 RUN  apt-get update && apt-get install -y libcurl4 && rm -rf /var/lib/apt/lists/*
+COPY --from=download /var/tmp/liftOver /usr/local/bin 
 COPY --from=download --chown=node:node /var/tmp/*.over.chain /home/node/
 
 # This layer is shared between the builder and production stages
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,5 +1,21 @@
+# This stage is tagged for local development.
+FROM node:16-slim AS dev
+WORKDIR /usr/src/app
+RUN apt-get update && apt-get install -y wget libcurl4
+RUN wget -O /usr/local/bin/liftOver https://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/liftOver  \
+  && chmod +x /usr/local/bin/liftOver
+COPY . .
+WORKDIR /home/node
+RUN mkdir chains && \
+  wget -O ./chains/hg19ToHg38.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz && \
+  wget -O ./chains/hg38ToHg19.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg38/liftOver/hg38ToHg19.over.chain.gz && \
+  gunzip ./chains/hg38ToHg19.over.chain.gz && \
+  gunzip ./chains/hg19ToHg38.over.chain.gz
+RUN chown -R node:node ./ && chown -R node:node /home/node
+WORKDIR /usr/src/app
+
 # This layer is shared between the builder and production stages
-FROM node:16-slim AS base
+FROM dev AS base
 WORKDIR /usr/src/app
 COPY package*.json yarn.lock ./
 RUN yarn --prod
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,23 +1,24 @@
-# This stage is tagged for local development.
+FROM curlimages/curl AS download
+WORKDIR /var/tmp
+RUN curl -O https://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/liftOver \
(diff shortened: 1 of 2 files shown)


server/Dockerfile
```diff
@@ -16,4 +32,4 @@ COPY --from=builder /usr/src/app/dist .
 ENV NODE_ENV=production
 USER node
 EXPOSE 3000
-CMD ["node", "."]
+CMD ["node", "."]
```
kevinlul: Keep trailing newline

server/Dockerfile
```diff
@@ -1,5 +1,21 @@
...
+  && chmod +x /usr/local/bin/liftOver
+COPY . .
+WORKDIR /home/node
+RUN mkdir chains && \
+  wget -O ./chains/hg19ToHg38.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz && \
+  wget -O ./chains/hg38ToHg19.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg38/liftOver/hg38ToHg19.over.chain.gz && \
+  gunzip ./chains/hg38ToHg19.over.chain.gz && \
+  gunzip ./chains/hg19ToHg38.over.chain.gz
+RUN chown -R node:node ./ && chown -R node:node /home/node
+WORKDIR /usr/src/app
```
kevinlul: This isn't great as a base build stage. I think you can have something like this, but use a `COPY` directive later on to grab the files that you need. For example, a `download` stage using `curlimages/curl`, download and unzip, and then later on in a different stage, do `COPY --from=download --chown node:node STUFF DESTINATION`

server/Dockerfile
```diff
@@ -1,23 +1,24 @@
-# This stage is tagged for local development.
+FROM curlimages/curl AS download
+WORKDIR /var/tmp
+RUN curl -O https://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/liftOver \
+    -O https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz \
+    -O https://hgdownload.cse.ucsc.edu/goldenpath/hg38/liftOver/hg38ToHg19.over.chain.gz && \
```
kevinlul: Should add the `-f` or `--fail` flag to curl in case the server returns a 4xx or 5xx, and `-L` or `--location` in case the server administrators decide to move the files and leave a redirect.

server/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
-RUN mkdir chains && \
-  wget -O ./chains/hg19ToHg38.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz && \
-  wget -O ./chains/hg38ToHg19.over.chain.gz https://hgdownload.cse.ucsc.edu/goldenpath/hg38/liftOver/hg38ToHg19.over.chain.gz && \
-  gunzip ./chains/hg38ToHg19.over.chain.gz && \
-  gunzip ./chains/hg19ToHg38.over.chain.gz
-RUN chown -R node:node ./ && chown -R node:node /home/node
-WORKDIR /usr/src/app
+COPY --from=download /var/tmp/liftOver /usr/local/bin 
+RUN chmod +x /usr/local/bin/liftOver && apt-get update && apt-get install -y libcurl4
+COPY --from=download --chown=node:node /var/tmp/*.over.chain /home/node/
```
kevinlul: You should use this dev stage as the base layer for the next one so these downloads are cached as the lowest layer we add.

server/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
 
 # This layer is shared between the builder and production stages
-FROM dev AS base
+FROM node:16-slim AS base
 WORKDIR /usr/src/app
 COPY package*.json yarn.lock ./
+COPY --from=download /var/tmp/liftOver /usr/local/bin
+RUN chmod +x /usr/local/bin/liftOver && apt-get update && apt-get install -y libcurl4
```
kevinlul: `--chmod` is also an option to the copy directive, and you shouldn't need libcurl here anymore then?

jennyziyi-xu: [https://stackoverflow.com/questions/67910547/why-docker-copy-doesnt-change-file-permissions-chmod](https://stackoverflow.com/questions/67910547/why-docker-copy-doesnt-change-file-permissions-chmod) 

`--chmod` option for `COPY` requires BUILDKIT, should we add `--chmod` and build with `DOCKER_BUILDKIT=1 docker-compose ........` ?

server/Dockerfile
```diff
@@ -1,5 +1,21 @@
...
+RUN curl -f -L -O https://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/liftOver \
+    -f -L -O https://hgdownload.cse.ucsc.edu/goldenpath/hg19/liftOver/hg19ToHg38.over.chain.gz \
+    -f -L -O https://hgdownload.cse.ucsc.edu/goldenpath/hg38/liftOver/hg38ToHg19.over.chain.gz && \
+    gunzip ./hg38ToHg19.over.chain.gz && \
+    gunzip ./hg19ToHg38.over.chain.gz && \
+    chmod +x ./liftOver
+
+# Development image 
+FROM node:16-slim AS dev
+WORKDIR /usr/src/app
```
kevinlul: This directive isn't needed as it doesn't affect the commands in this stage and it's in the `base` stage

server/Dockerfile
```diff
@@ -1,5 +1,21 @@
...
+COPY --from=download /var/tmp/liftOver /usr/local/bin 
+RUN  apt-get update && apt-get install -y libcurl4 && rm -rf /var/lib/apt/lists/*
```
kevinlul, line 12: Move this before the two `COPY` directives as the files to be copied would change with each other due to being copied from the same place. Add a comment that libcurl4 is a dependency of liftOver

kevinlul (approved):
LGTM!

PR #237, opened 2022-05-02
Some more cleanup

### LINKED ISSUE #229: Docker: Minor improvements
- [x] rename local stack test-node-1 to test-node
- [x] ssmp-server -> server
- [x] image name ssmp -> osmp
- [x] chmod +x on keycloak setup scripts

- [x] tsconfig.json
- [x] unnecessary secrets

kevinlul:
- ssmp -> osmp where possible (Docker image name)
- ssmp-server -> server (service name)
- test-node-1 -> test-node (service name, local development)
- docker-compose.dev.yaml -> docker-compose.staging.yaml
- secrets deprecated as they are always known: TEST_NODE_URL, TEST_DATA_DB_PORT, TEST_DATA_DB_HOST
- keycloak mounts reorganized, script is executable
- tsconfig: current conventions, strict mode, source maps, incremental compilation, resolve JSON modules

Closes #229


server/tsconfig.json
```diff
@@ -1,26 +1,24 @@
...
-    "target": "es6",
-    "module": "commonjs",
-    "lib": ["ES6", "ES2017", "ES2018", "ES2019", "ES2020"],
-    "sourceMap": false,
-    "outDir": "./dist",
-    "rootDir": "src",
     "moduleResolution": "node",
-    "removeComments": true,
-    "noImplicitAny": true,
-    "strictNullChecks": true,
```
hannahle, line 12: Wouldn't it be useful to have `strictNullChecks`? So we don't call methods on something that's potentially undefined?

server/tsconfig.json
```diff
@@ -1,26 +1,24 @@
...
-    "strictFunctionTypes": true,
```
hannahle, line 13: Same comment with `strictFunctionTypes`?

docker-compose.staging.yaml
```diff
@@ -32,8 +32,8 @@ services:
     image: ghcr.io/ccmbioinfo/osmp-test-node
     environment:
       TEST_DATA_DB:
-      TEST_DATA_DB_PORT: 
-      TEST_DATA_DB_HOST: 
+      TEST_DATA_DB_PORT: 3306
+      TEST_DATA_DB_HOST: mysql
```
hannahle: I believe this is `test-node-mysql` for staging

kevinlul:
Please check the [TSConfig docs](https://www.typescriptlang.org/tsconfig#strict) and the [base config](https://github.com/tsconfig/bases/blob/main/bases/node16.json). These are all already implied by strict mode and the checks were made more stringent by turning everything on.

.gitignore
```diff
@@ -1,11 +1,13 @@
 node_modules
 build
+dist
 
 .DS_Store
 .env
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+*.tsbuildinfo
```
hannahle, line 10: Btw, what generates this?

kevinlul: Incremental TypeScript builds. It's not actually super relevant to this setup because ts-node-dev is used.

docker-compose.yaml
```diff
@@ -107,6 +107,6 @@ services:
     ports:
       - ${KEYCLOAK_PORT}:8080
     volumes:
-      - ./keycloak/scripts:/usr/scripts
-      - ./keycloak/scripts/disable-theme-cache.cli:/opt/jboss/startup-scripts/disable-theme-cache.cli
+      - ./keycloak/bin:/usr/local/bin
+      - ./keycloak/startup-scripts:/opt/jboss/startup-scripts
       - ./keycloak/theme/ssmp:/opt/jboss/keycloak/themes/ssmp
```
hannahle: We might as well change the folder name from `ssmp` -> `osmp`

kevinlul: I was afraid that might mess with the theme for the realm actually in use.

hannahle: It should be fine. The only thing that would need to change with this is in the `osmp/keycloak/scripts/bootstrap-keycloak.sh`, we'd need to change the theme name from `ssmp` to `osmp`

docker-compose.staging.yaml
hannahle: > Please check the [TSConfig docs](https://www.typescriptlang.org/tsconfig#strict) and the [base config](https://github.com/tsconfig/bases/blob/main/bases/node16.json). These are all already implied by strict mode and the checks were made more stringent by turning everything on.

I see, thanks!

docker-compose.yaml
hannahle: Then, what I'd do is `docker-compose up --force-recreate` and then run the bootstrap script again

hannahle: We're actually not using the same sets of command for loading theme in staging, so it doesn't get affected there

jennyziyi-xu:
The environment variables for `test-node` are not set in`.env.sample`, running `docker-compose up -d --build` will result in errors.

jennyziyi-xu (approved):
LGTM! These are very useful changes to keep the code stack up-to-date.

PR #238, opened 2022-05-03
Resolve assemblyId in variant data queries.

jennyziyi-xu:
- For remote-test node, remove `assemblyId` in the request. 
- For g4rd node, `assemblyId` is a required field as specified in the sample request [here](https://github.com/ccmbioinfo/report-scripts/blob/master/docs/phenotips-api.md#matching-endpoint). Set `assemblyId` to be `GRCh37` because `g4rd` only contains data in assembly `GRCh37`. 
- Add `assemblyId` in `useFetchVariantsQuery.ts` in order to get the assembly of the returned variant data.

Reason for these modifications:
The desired behaviour is as follows. No matter which assembly the user specifies, we want to return **all variant data in all assemblies** and then perform liftOver if necessary. Thus, we do not want to send assemblyId to the data source nodes.

### DIFF
--- a/server/src/resolvers/getVariantsResolver/adapters/g4rdAdapter.ts
+++ b/server/src/resolvers/getVariantsResolver/adapters/g4rdAdapter.ts
@@ -73,6 +73,9 @@ const getG4rdNodeQuery = async ({
   const url = `${process.env.G4RD_URL}/rest/variants/match`;
   const { position, ...gene } = geneInput;
   variant.assemblyId = 'GRCh37';
+  // For g4rd node, assemblyId is a required field as specified in this sample request:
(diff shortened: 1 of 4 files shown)


server/src/resolvers/getVariantsResolver/adapters/g4rdAdapter.ts
```diff
@@ -72,6 +72,7 @@ const getG4rdNodeQuery = async ({
   }
   const url = `${process.env.G4RD_URL}/rest/variants/match`;
   const { position, ...gene } = geneInput;
+  variant.assemblyId = 'GRCh37';
```
hannahle, line 75: Can you add a comment here why we specify it as GRCh37 for future developers?

hannahle (approved):
LGTM!

PR #241, opened 2022-05-05
Bug fix: Add working directory for local dev

jennyziyi-xu:
The working directory is not specified for the `dev` image in `server/Dockerfile`, thus, `docker-compose up ssmp-server` will fail because the command `yarn start` in `docker-compose.yaml` could not find `package.json`

In this PR, `working_dir: /usr/src/app` is added before running the command `yarn start`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,6 +22,7 @@ services:
       context: server
       target: dev
     image: ghcr.io/ccmbioinfo/ssmp:dev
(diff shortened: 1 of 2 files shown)


kevinlul (commented):
Oops, this was the result of my last suggestion. You could also move the WORKDIR directive up to the dev stage of the Dockerfile

kevinlul (approved):
LGTM!

PR #242, opened 2022-05-06
React: Fix zygosity count for data returned from g4rd

jennyziyi-xu:
For the columns `het count` and `homo count`, the count was not calculated correctly. 

Currently, the rows are displayed correctly only for g4rd node.


hannahle (approved):
LGTM!

PR #244, opened 2022-05-06
LiftOver implementation 

jennyziyi-xu:
!!! Changes in Docker files are merged from `develop`, `docker-compose up -d --build` is necessary. 

liftOver
-  liftOver is implemented for both conversions `GRCh37` -> `GRCh38` and `GRCh38` -> `GRCh37`.
- CADD annotations for GRCh38 are added. gnomAD annotations for GRCh38 are not added and will be added later when it's loaded in Minio. 
- Unmapped variants if there are any are displayed, they are not annotated with CADD nor gnomAD. 

Frontend: 
- `assemblyIdCurrent` is added to GraphQL queries in order for the frontend to know the current assembly of a row of data. 
- Two columns `Current Assembly` and `Original Assembly` are added to the frontend.

### DIFF
--- a/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
+++ b/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
@@ -85,7 +85,7 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
     dataForAnnotation = liftoverResults.dataForAnnotation;
     annotationPosition = liftoverResults.annotationPosition;
   }
-  
+
   // Cadd annotations for data in user requested assemblyId
   let data: VariantQueryDataResult[] = [];
   const caddAnnotationsPromise = fetchCaddAnnotations(annotationPosition, assemblyId);
@@ -101,6 +101,7 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
   // gnomAD annotations TODO: gnomAD annotations for GRCh38 are not available yet.
   data = await annotateGnomad(data ?? dataForAnnotation);
 
+  // return unmapped variants if there's any
   if (unliftedVariants.length) {
     data = data.concat(unliftedVariants);
   }
--- a/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
+++ b/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
@@ -97,7 +97,9 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
   }
 
   // gnomAD annotations TODO: gnomAD annotations for GRCh38 are not available yet.
-  data = await annotateGnomad(data ?? dataForAnnotation);
+  if (assemblyId === 'GRCh37'){
+    data = await annotateGnomad(data ?? dataForAnnotation);
+  }
 
   // return unmapped variants if there's any
   if (unliftedVariants.length) {
--- a/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
+++ b/server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
@@ -13,7 +13,6 @@ import fetchCaddAnnotations from './utils/fetchCaddAnnotations';
 import annotateCadd from './utils/annotateCadd';
 import annotateGnomad from './utils/annotateGnomad';
 import liftover from './utils/liftOver';
-import resolveAssembly from './utils/resolveAssembly';
 import getG4rdNodeQuery from './adapters/g4rdAdapter';
 
 const getVariants = async (parent: any, args: QueryInput): Promise<CombinedVariantQueryResponse> =>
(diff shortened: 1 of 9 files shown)


server/src/resolvers/getVariantsResolver/utils/liftOver.ts
```diff
@@ -0,0 +1,93 @@
+import { AssemblyId, VariantQueryDataResult } from '../../../types';
+import resolveAssembly from '../utils/resolveAssembly';
+const fs = require('fs/promises');
+const tmpdir = require('os').tmpdir;
+const path = require('path');
+const promiseExec = require('util').promisify(require('child_process').exec);
+
+const createTmpFile = async () => {
+  const filename = Math.random().toString(36).slice(2);
```
kevinlul: You don't need this. `mkdtemp` is already putting this file in a unique directory.

server/src/resolvers/getVariantsResolver/utils/liftOver.ts
```diff
@@ -0,0 +1,93 @@
+import { AssemblyId, VariantQueryDataResult } from '../../../types';
+import resolveAssembly from '../utils/resolveAssembly';
+const fs = require('fs/promises');
+const tmpdir = require('os').tmpdir;
+const path = require('path');
+const promiseExec = require('util').promisify(require('child_process').exec);
```
kevinlul: I'm pretty sure because you're using TypeScript, these should be `import` statements instead, and let the compiler convert these to CJS `require`

server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
```diff
@@ -65,19 +66,45 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
     }
   });
 
-  // once variants are merged, handle annotations
-  const caddAannotations = settled.find(
+  // filter data that are not in user requested assemblyId
```
hannahle, line 68: Can we refactor the code for liftover to a separate file? This `getVariantsResolver` file is getting large.

server/src/resolvers/getVariantsResolver/utils/liftOver.ts
hannahle: Agreed! You can do 

`import * as fs from 'fs'` for example

server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
jennyziyi-xu: Are you referring to refactor code for "caddAnnotations"? There is a separate file for liftover already

server/src/resolvers/getVariantsResolver/utils/liftOver.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const createTmpFile = async () => {
+  const filename = Math.random().toString(36).slice(2);
+  const dir = await fs.mkdtemp(path.join(tmpdir(), 'liftover-'));
+  return path.join(dir, filename);
+};
+
+// Get start positions of lifted variants.
+const parseBedStart = (bed: String) =>
```
hannahle, line 13: We should add some string validation with regex. It might help others understand what exact string format this string needs to conform to.

server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
```diff
@@ -65,19 +66,45 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
...
+  const dataForLiftover = combinedResults.filter(v => v.variant.assemblyId !== assemblyId);
+  // filter data that are already in user requested assemlbyId
+  let dataForAnnotation = combinedResults.filter(v => {
+    if (resolveAssembly(v.variant.assemblyId) === assemblyId) {
```
hannahle: Should we standardize the `variant.assemblyId` in the `adapter` files, instead of the `resolver` files? This is to avoid repetitively calling `resolveAssembly` every time we need to standardize the assembly.

server/src/resolvers/getVariantsResolver/getVariantsResolver.ts
```diff
@@ -65,19 +66,45 @@ const resolveVariantQuery = async (args: QueryInput): Promise<CombinedVariantQue
...
+      v.variant.assemblyIdCurrent = assemblyId;
+      return true;
+    } else return false;
+  });
+  let unliftedVariants: VariantQueryDataResult[] = [];
+
+  // perform liftOver if needed
+  if (dataForLiftover.length) {
+    const liftoverResults = await liftover(dataForAnnotation, dataForLiftover, assemblyId);
+    unliftedVariants = liftoverResults.unliftedVariants;
```
hannahle: These 3 lines can be written as:

`{ unliftedVariants, dataForAnnotation, annotationPosition } = liftoverResults;`

server/src/resolvers/getVariantsResolver/utils/liftOver.ts
jennyziyi-xu: I am not sure where we would apply string validation with regex. The output of liftOver is 2 BED files: one for lifted variants, one for unlifted variants.  BED file has a specific format: [https://genome.ucsc.edu/FAQ/FAQformat.html](https://genome.ucsc.edu/FAQ/FAQformat.html).

The BED file for unlifted variants has one line of comment starting with "#" preceding each line of variant position, thus, these two functions `parseBedStart`, `parseBedEnd` filter out rows that start with "#".

hannahle (approved):
LGTM!!

PR #247, opened 2022-05-11
Bug: Fixed numerical number filters

hannahle:
- Fixed weird behaviour with numerical filter
- Added numerical filters for numeric columns

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -225,18 +225,21 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         id: 'ad',
                         Header: 'AD',
                         width: getColumnWidth('AD'),
+                        filter: 'between',
                     },
                     {
                         accessor: 'dp',
                         id: 'dp',
                         Header: 'DP',
                         width: getColumnWidth('DP'),
+                        filter: 'between',
                     },
                     {
                         accessor: 'gq',
                         id: 'gq',
                         Header: 'GQ',
                         width: getColumnWidth('GQ'),
+                        filter: 'between',
                     },
                     {
                         accessor: 'individualId',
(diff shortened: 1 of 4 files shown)


jennyziyi-xu:
Screen Shot 2022-05-11 at 4 58 57 PM
Screen Shot 2022-05-11 at 4 59 43 PM

We have "smaller than" sign in some cases and "smaller and equal to" sign in other cases. I wonder if we should make it a bit more consistent.

hannahle:
> Screen Shot 2022-05-11 at 4 58 57 PM
> 
> Screen Shot 2022-05-11 at 4 59 43 PM
> 
> We have "smaller than" sign in some cases and "smaller and equal to" sign in other cases. I wonder if we should make it a bit more consistent.

Good point. I'll change the icon

jennyziyi-xu:
I'm not sure if this original bug is only happening on my end. When the filter is expanded, if I try to type in a number, the expanded box will close on itself.

themisteryang:
> I'm not sure if this original bug is only happening on my end. When the filter is expanded, if I try to type in a number, the expanded box will close on itself.

I'm encountering this issue too:
1. Open a numeric filter popover.
2. Open the comparison sign dropdown.
3. With the comparison sign dropdown open, instead of selecting a comparison sign, click on the text input.
4. The entire filter popover closes.

Instead, maybe we should have the comparison type dropdown close (keeping the currently selected comparison sign) when the user clicks on the text input without selecting a comparison type.

hannahle:
> > I'm not sure if this original bug is only happening on my end. When the filter is expanded, if I try to type in a number, the expanded box will close on itself.
> 
> I'm encountering this issue too:
> 
> 1. Open a numeric filter popover.
> 2. Open the comparison sign dropdown.
> 3. With the comparison sign dropdown open, instead of selecting a comparison sign, click on the text input.
> 4. The entire filter popover closes.
> 
> Instead, maybe we should have the comparison type dropdown close (keeping the currently selected comparison sign) when the user clicks on the text input without selecting a comparison type.

Ah yes, I know why this is happening. This has to do with the fact that if you click anywhere outside of the dropdown, it closes the dropdown immediately (`useClickAway` hook). I'll look into how we can ignore certain clickable components when using this hook.

jennyziyi-xu (approved):
LGTM!

themisteryang (approved):

PR #248, opened 2022-05-11
Bug: Modified drop down to show sex names instead of codes

hannahle:


jennyziyi-xu:
The modification looks good to me! Since the sexPhenotype field is from Phenotips endpoint, I think we can test again when Phenotips staging instance is back up.

jennyziyi-xu (approved):
I am also fine if we'd like to merge this in now.

PR #249, opened 2022-05-12
React: Added burden count

hannahle:


react/src/components/Table/Table.tsx
```diff
@@ -220,6 +221,12 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                         Header: 'Zygosity',
                         width: getColumnWidth('Zygosity'),
                     },
+                    {
+                        accessor: 'burdenCount',
+                        id: 'burdenCount',
+                        Header: 'Burden Count',
+                        width: getColumnWidth('Burden Count'),
+                    },
```
themisteryang: Should we also add the numerical comparison filter to the "Burden Count" column?

hannahle: Good point!

PR #251, opened 2022-05-17
Filter CADD annotation based on consScore

jennyziyi-xu:
- Prioritize annotation of the most deleteriousness transcript by using the CADD variant annotation with the highest consScore. If multiple variant annotations have the same consScore, the first encountered variant annotation is reported. 
- The correctness of annotation filtering still needs to be verified. This PR is meant to bring the changes to the staging instance such that Magda and Matt could also see the transcript on OSMP after filtering and verify on their end whether the transcripts on OSMP correspond to the transcripts on G4RD phenotips.


hannahle (approved):
Made a small variable change. Otherwise LGTM!

PR #252, opened 2022-05-18
Add cdna, aaChange, spliceAI scores, CADD scores

jennyziyi-xu:
- `cDNA`, `aaChange` should be “NA” when no values are reported. 
- Add `PHRED` CADD scaled score. 
- Add `spliceAIScore` and `spliceAIType`. (All 4 spliceAI scores: SpliceAI_acc_gain, SpliceAI_acc_loss, SpliceAI_don_gain, SpliceAI_don_loss. Value range: 0-1. Report the max of these 4 scores. If the maximum score is 0 or NA, report 0 and NA for the score type. )


hannahle:
> * `cDNA`, `aaChange` should be “NA” when no values are reported.
> * Add `PHRED` CADD scaled score.
> * Add `spliceAIScore` and `spliceAIType`. (All 4 spliceAI scores: SpliceAI_acc_gain, SpliceAI_acc_loss, SpliceAI_don_gain, SpliceAI_don_loss. Value range: 0-1. Report the max of these 4 scores. If the maximum score is 0 or NA, report 0 and NA for the score type. )

Can you also add gnomad_af = 0 if there's no value returned?

jennyziyi-xu:
Yes, I have formatted empty fields to be 0. I haven't changed the header name from `gnomad_exome_af` to `gnomad_af` and I think we could change that once we have changed the fields returned from gnomAD annotations.

hannahle (approved):

PR #254, opened 2022-05-26
React: Cache previously selected visible columns 

jennyziyi-xu:
- This PR only addresses issues within a gene search. If a second gene search is performed, the default visible columns is set to be all the columns in the group "Variant". However, if we want to conserve visible columns between gene searches, this issue will be addressed in another PR.  

- In the result table, collapsing a group will make all columns in the group invisible. Expanding a group will show only previously visible groups. 
- In the "Column VIsibility" pop-up, unchecking a group will uncheck all the columns in the group. Checking a group will only show cached visible columns ( in the case that there are no cached visible columns, all the columns in the group will be shown). Closing the "Column Visibility" pop-up will not save the changes made in the pop-up. 
- In summary, there is a cache `cacheTable` for the result table and a cache `cacheVisibility` for the "Column Visibility" pop-up. Every time user opens the "Column Visibility" modal, `cacheVisibility` is initialized to be equal to `cacheTable`. Every time user clicks `apply` in the modal, `cacheTable` is updated to be equal to `cacheVisibility`.

### DIFF
--- a/react/src/components/Table/ColumnVisibilityModal.tsx
+++ b/react/src/components/Table/ColumnVisibilityModal.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { DragDropContext, Draggable, Droppable, DropResult } from 'react-beautiful-dnd';
 import { BsFillEyeFill, BsFillEyeSlashFill } from 'react-icons/bs';
 import { ColumnInstance, HeaderGroup, UseTableInstanceProps } from 'react-table';
@@ -9,6 +9,7 @@ interface ColumnVisibilityModalProps<T extends object>
     extends Pick<UseTableInstanceProps<T>, 'toggleHideColumn'> {
     headerGroups: HeaderGroup<T>[];
     allColumns: ColumnInstance<T>[];
+    visibleColumns: ColumnInstance<T>[];
     setColumnOrder: (update: string[] | ((columnOrder: string[]) => string[])) => void;
     cached: Record<string, boolean>;
     setCached: (value: Record<string, boolean>) => void;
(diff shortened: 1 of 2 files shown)


react/src/components/Table/ColumnVisibilityModal.tsx
```diff
@@ -1,34 +1,74 @@
...
+        return !checkedColumns[groupMapping[header]];
+    };
+
+    // Find a column's group: "Variant", "Variant Details" or "Case Details".
+    const columnToGroup = (columnId: string) => {
+        let group = headerGroups[0].headers[0];
+        headerGroups[0].headers.forEach(header => {
+            if (header.columns?.find(c => c.id === columnId)) group = header;
+        });
+        return group;
```
kevinlul: So this will default to the first column?

jennyziyi-xu: Yes, since each column belongs to a group, the default group should not cause any issues. 

It's better to remove this default group though, and I just removed it.

themisteryang (approved):
<details>
  <summary>LGTM, the bug looks to be fully fixed!</summary>
  Just for reference, the way to reproduce the bug was:
  
  1. Start (by default) with all the columns checked.
  2. Uncheck `Chr` and `Start`.
  3. Uncheck `Variant` (this unchecks all of the `Variant` columns, as expected).
  4. Check `Chr` (this also checks `Variant`, as expected).
  5. Uncheck `Variant` again (this unchecks all of the `Variant` columns, as expected).
  6. Check `Variant`.
  
  Now, every `Variant` column except for `Start` is checked instead of only `Chr` being checked.
</details>

PR #257, opened 2022-05-31
React: Show "This patient is clinically normal" 

### LINKED ISSUE #256: React: "Phenotypes" column width
Screen Shot 2022-05-31 at 12 11 58 PM

- When a user expands the `phenotype` terms, the terms are not wrapped according to the width of the column. We would like all text should be visible to the user. 

- We can also increase the width of the column `phenotypes` by setting a fixed value. Or we can adjust this column's width dynamically according to the length of the data.

jennyziyi-xu:
1, According to the [OSMP Gantt chart "OSMPColumnsMapping" tab](https://docs.google.com/spreadsheets/d/1eztLBGeqjYj3IafpPxhZo5mbvz6Xzw0LbP_9_HpxeHA/edit#gid=2130371978),
for the column "Phenotypes",

- If the features list in the patient response is not empty, this means there are phenotypes. Show the list of phenotypes on OSMP.
- If the feature list is empty, check `clinicalStatus`. If `clinicalStatus = "unaffected"`, show "This patient is clinically normal".

2, Remove `datasetID` from GraphQL 
3, CSS styling changes in table cells for text wrapping.  
4, Make the column `contactInfo` wider to fully display all the content.

### DIFF
--- a/react/src/apollo/hooks/useFetchVariantsQuery.ts
+++ b/react/src/apollo/hooks/useFetchVariantsQuery.ts
@@ -43,7 +43,6 @@ const fetchVariantsQuery = gql`
                     variantId
                 }
                 individual {
-                    datasetId
                     diseases {
                         ageOfOnset {
                             age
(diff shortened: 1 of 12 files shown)


themisteryang (approved):
I added two minor fixes, everything else LGTM!

PR #258, opened 2022-06-01
Keycloak: script for adding new users based on list of emails

### LINKED ISSUE #245: Keycloak: Develop script to automatically add G4RD users to Keycloak based on list of emails

frewmack:
Closes #245 

EDIT:
`example.txt`:

```txt
user-alpha@domain.ca
user-bravo@domain.ca

```

Usage:

```
cat example.txt | admin-scripts/add-users-by-email.sh
```


frewmack:
Marking this branch as draft. Realized that user IDs need to be pulled from Auth0 for G4RD users, and this isn't implemented here yet.

.env.sample
```diff
@@ -45,3 +45,4 @@ G4RD_CLIENT_ID=exampleclientid
 G4RD_GRANT_TYPE=granttype
 G4RD_AUTH_METHOD=bearer
 G4RD_TOKEN_URL=https://some-provider.example.com
+G4RD_AUTH0_BASE_URL=https://some-subdomain.auth0.com
```
kevinlul: Trailing newline

keycloak/bin/add-users-by-email.sh
```diff
@@ -0,0 +1,71 @@
...
+        --header "Content-Type: application/x-www-form-urlencoded" \
+        --data "client_id=${AUTH0_BROKER_CLIENT_ID}" \
+        --data "grant_type=client_credentials" \
+        --data "client_secret=${AUTH0_BROKER_CLIENT_SECRET}" \
+        --data "audience=${AUTH0_BROKER_MANAGEMENT_API_URL}"
+    )
+fi
+
+accesstoken=$(
+    echo "${tokenresponse}" | grep -zoP '"access_token":\s*"\K[^\s,]*(?="\s*,)'
```
kevinlul: Are you trying to parse JSON with grep?

frewmack: Yep. The regex matches the value inside the double quotes whose key is "access_token". Using something like [jq](https://stedolan.github.io/jq/) would probably be better, but it's not included in the keycloak container by default.

kevinlul: This script bin is only mounted in development. Is this a script to be used in development, or run from a development machine? If so, we can write this script to assume `jq` exists (heck, even Node.js) and get more reliable parsing there, and use `docker-compose exec keycloak` for the calls to kcadm.

frewmack: My understanding was that this script would be used to semi-automate the process of adding G4RD users to OSMP Keycloak in production. I suppose it does make more sense then to fix the script to run outside of Docker by some trusted admin user in production. I'll rework the script.

admin-scripts/add-users-by-email.sh
```diff
@@ -0,0 +1,66 @@
...
+        --data "audience=${G4RD_AUTH0_BASE_URL}api/v2/"
+    )
+fi
+
+accesstoken=$(
+    echo "${tokenresponse}" | jq -r '."access_token"'
+)
+
+# Get users by email
+file=${1:--}
```
kevinlul:
```suggestion
file=${1:-/dev/stdin}
```

admin-scripts/add-users-by-email.sh
```diff
@@ -0,0 +1,66 @@
...
+    printf "Adding user with email '%s' and id '%s'... " "$line" "$userid"
+    set +e
+    docker-compose exec keycloak /opt/jboss/keycloak/bin/kcadm.sh create users -s username="$userid" -s email="$line" -s enabled=true -r "${KEYCLOAK_REALM}"
+    if [[ $? -eq 0 ]]; then
+        printf "User added successfully\n"
+    else
+        printf "Failed to add user\n"
+    fi
+    set -e
+done < <(cat -- "$file")
```
kevinlul:
```suggestion
done < "$file"
```
Avoids starting a subshell just to read a file or standard input

admin-scripts/add-users-by-email.sh
```diff
@@ -0,0 +1,66 @@
...
+            --url "${G4RD_AUTH0_BASE_URL}api/v2/users-by-email?email=${line}" \
+            --header "Authorization: Bearer $accesstoken"
+    )
+    userid=$(echo "$userobj" | jq -r '.[]["user_id"]')
+
+    # try to add user to keycloak with this id
+    printf "Adding user with email '%s' and id '%s'... " "$line" "$userid"
+    set +e
+    docker-compose exec keycloak /opt/jboss/keycloak/bin/kcadm.sh create users -s username="$userid" -s email="$line" -s enabled=true -r "${KEYCLOAK_REALM}"
+    if [[ $? -eq 0 ]]; then
```
kevinlul: I believe you can also just do `if command-here` but this is probably more clear

jennyziyi-xu:
Screen Shot 2022-07-14 at 4 40 05 PM

Hi Mackenzie, this is my example.txt, I was able to run your script and the user with the first email was added to Keycloak successfully! However, the second email was not processed. If I put the second email on the first line, it will be processed. Do I have the correct format for example.txt? It seems only the email on the first line is processed.

jennyziyi-xu:
I think `G4RD_CLIENT_SECRET` is missing from `.env.sample`.

jennyziyi-xu (approved):
LGTM!!!

PR #262, opened 2022-06-14
React: UI changes

themisteryang:
- Add numeric filter to the "Homo Count" and "Het Count" columns
- Add a more distinct divider between the different sets of columns (i.e. the "Variant", "Variant Details", and "Case Details" columns)
- Move the contributors selection and replace the checkboxes with a multi-selection select
- Increase the screen real estate for the variant query table by condensing various UI elements and reducing margins and paddings wherever possible
- Modify UI behavior so that changing the "Genome Assembly" no longer clears the "Gene Name" input


themisteryang:
**Before:**
Screen Shot 2022-06-14 at 10 27 31 AM

**After:**
Screen Shot 2022-06-14 at 10 27 40 AM

jennyziyi-xu:
I like the new user interface! For some numerical filters (for example, "Homo count" and "Het count"), filtering for "value = 0" doesn't work.

jennyziyi-xu:
> I like the new user interface! For some numerical filters (for example, "Homo count" and "Het count"), filtering for "value = 0" doesn't work.

Since this is an issue that existed before this pull request, we could fix this issue in another PR because this PR already contains a lot of changes.

themisteryang:
> > I like the new user interface! For some numerical filters (for example, "Homo count" and "Het count"), filtering for "value = 0" doesn't work.
> 
> Since this is an issue that existed before this pull request, we could fix this issue in another PR because this PR already contains a lot of changes.

I've gone ahead and added the bugfix to this PR since it was just a one-line change!

jennyziyi-xu:
required-fields
mandatory-fields

To further clarify that all the fields are required, I think we can add the word "required" to all the fields or add the phrase "all the fields are mandatory".

themisteryang:
> required-fields mandatory-fields
> 
> To further clarify that all the fields are required, I think we can add the word "required" to all the fields or add the phrase "all the fields are mandatory".

Good point, just having an asterisk might not be clear enough, I'll add a "Required" indicator to the fields!

jennyziyi-xu:
The "Required" textbox looks a bit bulky, I think we can go with this format: 
mandatory-fields

For each input field, we can display red asterisks and remove "Required" textboxes. 
In the input search area, we can display "* Mandatory fields" in red.

frewmack:
The filter operator button for some numeric filters has some strange display behaviour:
image

This seems to be due to the operator dropdown being rendered inside the textbox div instead of as a separate floating element. This already existed before this PR so we can leave it as a separate issue.

frewmack (approved):
Looks good to me!

themisteryang:
> The "Required" textbox looks a bit bulky, I think we can go with this format: mandatory-fields
> 
> For each input field, we can display red asterisks and remove "Required" textboxes. In the input search area, we can display "* Mandatory fields" in red.

I've simplified the indicator used for required fields:
Screen Shot 2022-06-15 at 11 30 30 AM

jennyziyi-xu:
I like this layout, thanks very much for making the change!

jennyziyi-xu (approved):
LGTM!

PR #264, opened 2022-06-14
React: Fix removal of non-unique variants for remote-test node 

### LINKED ISSUE #243: React: Bug in remote-test node display.
If a user queries from remote-test node, the results are not displayed correctly: 1) When `Case Details` section is collapsed, non-unique variants are displayed. 2) Row alternating coloring does not work as expected. 

This is because the field `individual ID` is `null` for the payload from remote-test node.

frewmack:
Closes #243 
Table rows are sorted by `uniqueId` after we get the `uniqueVariantIndices` from `prepareData`, meaning that they can fall out-of-sync since the order is not preserved. The change uses only the `uniqueId` of the current and previous rows to determine whether to show the row or not when the Case Details columns are hidden.

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -644,7 +644,7 @@ const Table: React.FC<TableProps> = ({ variantData }) => {
                                                                     justifyContent="center"
                                                                     nowrap
                                                                 >
-                                                                    {/* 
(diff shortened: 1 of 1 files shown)


jennyziyi-xu (approved):
LGTM! Turns out it's a simple fix!

PR #265, opened 2022-06-14
Bug: Remove "transcript" field from Phenotips endpoint. 

### LINKED ISSUE #260: Bug: transcript and aaChange should come from CaddAnnotations instead of PT endpoint

jennyziyi-xu:
1, Before, if CADD annotation is not performed (it should be performed unless there is an expected backend error or the file is removed from the remote URL), the field `transcript` would be still returned from Phenotips endpoint and be displayed. If CADD annotation is not performed, the column `transcript` should be empty. 

In this PR, `transcript` is removed from the Phenotips endpoint's payload.

2, Code clean-up: change `referenceName` to `chromosome`.


frewmack (approved):
LGTM!

PR #266, opened 2022-06-15
React: Modify "Affected Status" and "Phenotypes" columns

jennyziyi-xu:
According to Magda and Matt's feedback, see [OSMP Gantt chart](https://docs.google.com/spreadsheets/d/1eztLBGeqjYj3IafpPxhZo5mbvz6Xzw0LbP_9_HpxeHA/edit#gid=2130371978)

- For "Affected Status" column, we should display "affected" or "unaffected" according to `clinicalStatus` returned from the Phenotips endpoint. 
- For "Phenotypes" column, we should remove the display of "This patient is clinically normal" and display phenotypes when they exist.


frewmack (approved):
LGTM! 👍

themisteryang (approved):
LGTM!

PR #277, opened 2022-06-17
Bug: Fix a bug in the Button component and the numeric filter

themisteryang:
- Fix an infinite render loop (maximum update depth exceeded) bug in the Button component caused by setState calls in the effect hook
- Fix a numeric filter bug in which negative array indexing (which returns undefined) causes the access of nested object properties to fail/error


frewmack (approved):
LGTM!

PR #279, opened 2022-06-17
Bug: Fix a bug with the column group divider

themisteryang:
Fix a bug with the column group divider where it can become misaligned/out of sync if a column is reordered to be the last in its set


frewmack (approved):
Works on my end!

jennyziyi-xu:
Screen Shot 2022-06-17 at 2 17 47 PM

Thanks for the fix, Kevin! I think there is still a little bug for the last group "Case Details". This screenshot shows the result table after selecting two columns in the last group and changing their order.

jennyziyi-xu (approved):
LGTM, thanks for the quick fix!

PR #280, opened 2022-06-17
DevOps: Github workflow jobs should fail on yarn check-format

### LINKED ISSUE #278: Github: Github actions should fail when files are not formatted with "yarn format"
Currently in a pull request, even though some files need to be formatted with "yarn format", the jobs `build-react` and `build-server` would still be completed successfully. 

- [ ] The corresponding job should fail when there are files that need to be formatted with "yarn format".

jennyziyi-xu:
Closes #278 

- Changes are made in `node.yml` and `react.yml` such that the Github workflow jobs will fail if there are files that need to be formatted with `yarn format`. 
- `yarn format` is run for all the files in `react`, `server` and `test-node` because the recent commit "React UI changes" to the `develop` branch contains unformatted files. 

- In the future, `yarn format` needs to be run prior to pushing a commit to a PR in order for all checks to pass.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -87,8 +87,7 @@ jobs:
         continue-on-error: true
         run: yarn lint
       - if: github.event_name == 'pull_request'
-        continue-on-error: true
-        run: yarn format
+        run: yarn check-format
       - uses: docker/login-action@v1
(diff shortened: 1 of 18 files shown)


themisteryang (approved):
LGTM!

PR #283, opened 2022-06-22
React Apollo: Use non-subscription approach for remote node error link

### LINKED ISSUE #103: React: Handle 403s
Currently a `403` response from the `/graphql` endpoint hangs until timeout. We should do some basic handling (redirect to login) and in the meantime look into why refresh tokens may be going stale.

frewmack:
Closes #103 ~~(partially)~~

Definitely need someone with more experience with Apollo to validate this.

My understanding of the issue is that previously, `remoteNodeErrorLink` used a subscription (ie. a WebSocket connection) to check for gql errors. This link was suppressing network errors for some reason, causing failed gql requests to hang until the timeout link triggered. I've changed this to a link that does not use subscriptions and it appears to work fine. 

~~I'll create a new issue for redirecting to login on 403.~~ Login redirects should happen automatically, so I will update this PR to include a prompt to refresh the page.

### DIFF
--- a/react/src/apollo/client.ts
+++ b/react/src/apollo/client.ts
@@ -1,6 +1,6 @@
 import { ApolloClient, createHttpLink, from, InMemoryCache, useLazyQuery } from '@apollo/client';
 import { onError } from '@apollo/client/link/error';
-import { ApolloLink, QueryHookOptions, useQuery } from '@apollo/react-hooks';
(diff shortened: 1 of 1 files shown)


jennyziyi-xu (approved):

jennyziyi-xu:
Verified that network error 404 pops up immediately whereas on the `develop` branch, it will hang until timeout.

PR #284, opened 2022-06-27
Show 422 error when CADD annotations are not performed

### LINKED ISSUE #271: React: Display a message when CADD annotations are not performed
- [ ]  In `fetchCaddAnnotations.ts`, if a gene is larger than 600,000 base pairs, CADD annotation is not performed. In this case, the frontend should display a message indicating that CADD annotation has not been performed. For example, "Gene of size 700,000 bp is too large to annotate with VEP. Annotating with gnomAD only!"

- This is the desired behavior because the server does not have enough RAM to handle the subsequent matching between variant data and CADD variant annotation. Eventually, when the HPF Slurm automation is ready, CADD annotation will be performed for all genes even if they are larger than 600,000 base pairs.

frewmack:
Closes #271

### DIFF
--- a/react/src/apollo/client.ts
+++ b/react/src/apollo/client.ts
@@ -24,10 +24,15 @@ export const buildLink = (token?: string) => {
         return forward(operation).map(result => {
             // https://github.com/apollographql/apollo-link/issues/298
             const errorDispatch = operation.getContext().dispatch;
-            // const errorDispatch = result.context!.dispatch;  // Is this more 'correct'?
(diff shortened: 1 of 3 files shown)


jennyziyi-xu:
Should we display a simple human-readable error message such as "Error in performing CADD annotations: skipping CADD annotations" instead of the Promise error? 
Screen Shot 2022-06-30 at 11 55 56 AM

jennyziyi-xu (approved):
LGTM!

PR #286, opened 2022-06-28
React: Remove pagination from table; fix cached results not updating table correctly

### LINKED ISSUE #285: React: Cached search results do not update table size correctly
If a gene search is displayed from the cache after a gene search with fewer variants, the table will not increase in size to display all the unique variants of the cached gene search.

**Steps to reproduce:**
1. Search for a gene with a larger number of unique variants (eg. ABHD16A on remote-test, 26 variants)
2. Search for a gene with a smaller number of unique variants than step 1 (eg. FAIM on remote-test, 4 variants)
3. Search for the same gene from step 1 again. The cached result will populate the table, however there will only be as many rows as there were for step 2, despite there being more variants to display.

frewmack:
Closes #285 

Retrieving search results from cache updates the table correctly, but not the size of the page. Since we only ever display *all* results in the table with no pagination controls, I removed any mention of pagination from the table and opt to read from `tableInstance.rows` instead of `tableInstance.page`.

### DIFF
--- a/react/src/components/Table/Table.tsx
+++ b/react/src/components/Table/Table.tsx
@@ -15,7 +15,6 @@ import {
     useFilters,
     useFlexLayout,
     useGlobalFilter,
-    usePagination,
     useResizeColumns,
     useSortBy,
     useTable,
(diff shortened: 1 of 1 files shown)


jennyziyi-xu (approved):
LGTM! Nice bug catch!

PR #289, opened 2022-07-05
React: Add scroll to bottom on Table render

frewmack:
Small UX change; scroll to bottom of page automatically when Table renders


jennyziyi-xu (approved):
LGTM!

PR #295, opened 2022-07-25
Bug: Fix a bug with the column group divider

themisteryang:
Fix a bug that creates an incorrect column group divider if a column's right-adjacent column is hidden


jennyziyi-xu (approved):
LGTM!

PR #302, opened 2022-08-03
"Deploy" job only runs when the workflow is against a specific branch

jennyziyi-xu:
The "Deploy" stage should run when there is a push to `develop` or `production` and when we manually run a worflow. Thus, the  "Deploy" stage's trigger condition has been changed from "on push event" to "on which branch is targeted".


frewmack (approved):
Looks sound! 🚀

themisteryang (approved):
Everything checks out! 🚀

PR #306, opened 2022-08-05
Server: Fix memory leaks in express-session 

jennyziyi-xu:
When `server` container is deployed to production, this error is encountered: 
"Warning The default server-side session storage, MemoryStore, is purposely not designed for a production environment. It will leak memory under most conditions, does not scale past a single process, and is meant for debugging and developing."

Among a list of compatible session stores for `express-session`, I have chosen [`memorystore`](https://www.npmjs.com/package/memorystore)  for its ability to avoid memory leaks.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -16,7 +16,7 @@ import memorystore from 'memorystore';
 
 const app = express();
 
-const memoryStore = memorystore(session);
+const MemoryStore = memorystore(session);
 
 mongoose
   .connect(process.env.MONGO_CONNECTION_STRING!)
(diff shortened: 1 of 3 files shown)


server/src/server.ts
```diff
@@ -32,13 +33,15 @@ app.use(
...
+    store: new MemoryStore({
+      checkPeriod: 86400000, // prune expired entries every 24h
+    }),
   })
 );
 
 const keycloak = new Keycloak(
   {
-    store: memoryStore,
+    store: MemoryStore,
```
frewmack: This is the memory store constructor, not the object. Both the session and Keycloak should use the same memory store object* as before.

frewmack (changes requested):

frewmack (changes requested):
LGTM!

frewmack (approved):

PR #311, opened 2022-08-12
React: Show loading UI when refetching queries

### LINKED ISSUE #310: React: Re-submitting query more than once does not show loading state
Sending an identical query two times in a row correctly sends the second query, however it does not provide any indication to the user that the second query is in progress.

frewmack:
Closes #310


jennyziyi-xu (approved):
LGTM!

PR #314, opened 2022-08-16
React / API: Add "observed" field for phenotypic features

### LINKED ISSUE #312: React / API: Mark phenotypes with "NO" if they are present in the query but not present in the participant
Currently, any phenotype that appears for a particular case is included in the "Phenotypes" column on the frontend. However, in PhenoTips, there are sometimes phenotypes that are marked explicitly as not being present (eg. "Short Stature", "NO Tall Stature"), yet these are listed the same way in OSMP (eg. "Short Stature", "Tall Stature"). The frontend should show phenotypes that are marked with "NO" the same way they are presented in PhenoTips.

https://docs.phenotips.com/reference/getpatient-1 Marked in the "observed" field.

frewmack:
Closes #312 

Frontend now shows "YES X" or "NO X" for Phenotypes to indicate whether a phenotype was observed by a clinician.

Based on the results from PhenoTips, the logic is set up as follows:
```
observed field = "yes"  => true
observed field = "no"   => false
observed field missing  => true
```
We may need to verify with collaborators before pushing this to `develop`.

### DIFF
--- a/react/src/apollo/hooks/useFetchVariantsQuery.ts
+++ b/react/src/apollo/hooks/useFetchVariantsQuery.ts
@@ -78,6 +78,7 @@ const fetchVariantsQuery = gql`
                         onsetType
                         phenotypeId
                         phenotypeLabel
+                        observed
                     }
                     sex
                 }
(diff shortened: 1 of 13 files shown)


jennyziyi-xu (approved):
LGTM

themisteryang (approved):
LGTM! 🚀

PR #315, opened 2022-08-17
Bug: Fix a bug that prevented certain variants from being annotated with gnomAD

themisteryang:
Fix a bug where certain variants are not being annotated with gnomAD if their chromosome value is prefixed with `chr`


jennyziyi-xu (approved):
LGTM

jennyziyi-xu (approved):

PR #316, opened 2022-08-17
React: Allow users to fetch additional gene hits beyond the current default of 10

### LINKED ISSUE #313: React: Only the first ten matching gene hits are being returned from the mygene.info gene query service
Currently, for the gene search autocomplete, OSMP is only querying for the first ten matching gene hits ([as the size parameter defaults to 10](https://docs.mygene.info/en/latest/doc/query_service.html#size)). As a result, any additional gene hits beyond the first ten are not being returned to OSMP. Instead, we want to allow the user to fetch more gene hits (if available).

themisteryang:
Currently, for the gene search autocomplete, OSMP is only querying for the first ten matching gene hits from the [mygene.info gene query service](https://docs.mygene.info/en/latest/doc/query_service.html#get-request) (the default when no value is provided to the `size` query parameter) with no option to allow the user to fetch additional gene hits.

This PR addresses this issue by adding a button to the gene autocomplete selectable list that allows users to fetch ten additional gene hits at a time.

**When there are additional gene hits to fetch:**
Screen Shot 2022-08-17 at 3 17 09 PM

**When all genes hits have been fetched:**
Screen Shot 2022-08-17 at 3 18 50 PM

Closes #313

### DIFF
--- a/react/src/apollo/client.ts
+++ b/react/src/apollo/client.ts
@@ -1,5 +1,6 @@
 import { ApolloClient, createHttpLink, from, InMemoryCache, useLazyQuery } from '@apollo/client';
 import { onError } from '@apollo/client/link/error';
+import { relayStylePagination } from '@apollo/client/utilities';
 import { ApolloLink, QueryHookOptions, ServerError, useQuery } from '@apollo/react-hooks';
 import { RestLink } from 'apollo-link-rest';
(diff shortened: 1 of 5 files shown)


frewmack (approved):
LGTM!

jennyziyi-xu (approved):

